PR #5, opened 2019-06-20
start providing 0.4.1/0.4.2 api and fix some compliance issues

soxofaan:


soxofaan:
(fixed conflict with 4447827)

Issue #8, opened 2019-07-16
Encapsulate "driver implementation" functions


soxofaan:
openeo-python-driver allows a plugable "DRIVER_IMPLEMENTATION_PACKAGE", but does this quite rudimentary at the moment. E.g:

```
i = importlib.import_module(os.getenv('DRIVER_IMPLEMENTATION_PACKAGE', "dummy_impl"))
getImageCollection = i.getImageCollection
get_layers = i.get_layers
try:
    create_process_visitor = i.create_process_visitor
except AttributeError as e:
    create_process_visitor = None
create_batch_job = i.create_batch_job
run_batch_job = i.run_batch_job
get_batch_job_info = i.get_batch_job_info
get_batch_job_result_filenames = i.get_batch_job_result_filenames
```
( https://github.com/Open-EO/openeo-python-driver/blob/314023137/openeo_driver/ProcessGraphDeserializer.py#L67 )

This approach of copy-pasting  toplevel functions scales quite ugly and ruins code intelligence in your editor/IDE.

It would be cleaner to encapsulate this intermediate "API" e.g. as (abstract) classes that backends have to implement. This is more explicit API-wise, more self documenting and allows your IDE to be more helpful.

soxofaan:
done (as part of #34)

PR #10, opened 2019-07-17
Refactor secondary service management and allow service deletion (python-driver part)

soxofaan:
This is quite big PR to enable deletion of secondary services (the openeo-python-driver part). The deletion part is not that big on its own, but I took the opportunity to refactor some parts to make things more encapsulated and self-documenting (also see #8 ).

Instead of copy-pasting a bunch of top-level functions into `openeo_driver/ProcessGraphDeserializer.py`,  which is ugly and not very developer friendly  I created a new `openeo_driver/backend.py` module intended to contain base classes that describe what openeo_driver expects from an actual backend. This enables a cleaner decoupling of the flask REST part from backend implementation details with a more explicit (type hinted) API. I only started with the "secondary services" microservice here, but over time we should port more this way.

Also includes some unit test fixes and type hint additions along the way.

This PR will be accompanied with a related PR on openeo-geopyspark-driver , which have to be merged in tandem because the changes are quite intertwined. I don't think it is very useful in this phase of these projects to decouple these PRs and provide some kind of backward/forward compatibility mechanism


soxofaan:
just created the geopyspark cousin PR: Open-EO/openeo-geopyspark-driver#13

PR #27, opened 2020-01-27
EP-3236 use geopandas for total bounds calculation in DelayedVector

soxofaan:
The original implementation scales very bad for large number of geometries (roughly quadratic)
This new geopandas based implementation scales less than linear


soxofaan:
indeed, geopandas was already in requirements.txt, but not in setup.py

PR #37, opened 2020-04-24
Add support for "short version" url style: e.g. `/openeo/0.4/` and `openeo/1.0/`

soxofaan:
While diving into openeo deploy pipelines I played with this:
support for short version URLS (only MAJOR.MINOR), e.g. `/openeo/0.4/` which maps to api version 0.4.2 internally.
That way users don't have to worry about PATCH component


jdries:
Nice one, I was already wondering how to deal with this. Thanks!

PR #39, opened 2020-05-19
Feature/EP-3276

laxiwuka:
Changes to comply with 1.0 spec

### DIFF
--- a/openeo_driver/views.py
+++ b/openeo_driver/views.py
@@ -653,8 +653,9 @@ def _jsonable_service_metadata(metadata: ServiceMetadata, full=True) -> dict:
 @api_endpoint
 @openeo_bp.route('/services', methods=['GET'])
 @auth_handler.requires_bearer_auth
-def services_get():
+def services_get(user: User):
     """List all running secondary web services for authenticated user"""
+    # TODO implement user level secondary service management EP-3411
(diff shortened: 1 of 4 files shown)


openeo_driver/views.py
```diff
@@ -648,9 +652,9 @@ def _jsonable_service_metadata(metadata: ServiceMetadata, full=True) -> dict:
 
 @api_endpoint
 @openeo_bp.route('/services', methods=['GET'])
+@auth_handler.requires_bearer_auth
```
soxofaan, line 655: in addition to `requires_bearer_auth`, can you add a `user: User` argument to these view functions and a comment like `TODO implement user level secondary service management EP-3411`

tests/test_views.py
```diff
@@ -659,7 +663,7 @@ def test_list_services_040(self):
         }
 
     def test_list_services_100(self):
-        metadata = self.client.get('/openeo/1.0.0/services').json
+        metadata = self.client.get('/openeo/1.0.0/services', headers=self._auth_header).json
```
soxofaan, line 666: maybe also add tests to check that authentication is required for a couple of endpoints

soxofaan:
looks good overall

PR #63, opened 2021-04-13
Download results with signed url's

laxiwuka:


openeo_driver/errors.py
```diff
@@ -215,6 +215,17 @@ def __init__(self, filename: str):
         super().__init__(message=self.message.format(file=filename))
 
 
+class FileExpiredException(OpenEOApiException):
```
soxofaan, line 218: New errors in this file have first to be defined through https://github.com/Open-EO/openeo-api/blob/master/errors.json

see https://github.com/Open-EO/openeo-api/issues/379

soxofaan: in the mean time, you can just raise a generic exception:

```python
        raise OpenEOApiException(
            status_code=410,
            code="FileExpired",
            message="File '{file}' has expired".format(file=filename)"
         )
```

soxofaan: Interesting suggestion from https://github.com/Open-EO/openeo-api/issues/379#issuecomment-819456415:
raise it with something like:

```python
raise OpenEOApiException( 
    status_code=410, 
    code="AssetLinkExpired", 
    message="Batch job asset link has expired. Request the batch job results again for fresh asset links" ,
)
```

soxofaan:
FYI: Open-EO/openeo-api#380 is now merged in draft and the newly introduced error is 
```
	"ResultLinkExpired": {
		"description": "The signed URLs for batch job results have expired. Please send a request to `GET /jobs/{job_id}/results` to refresh the links.",
		"message": "The link to the batch job result has expired. Please request the results again.",
		"http": 410,
```
so you could already adapt that error code `ResultLinkExpired` too

soxofaan:
is that bump of the openeo_driver/specs/openeo-api/0.4 submodule intentional?

PR #64, opened 2021-04-13
Support feature collection in read vector

laxiwuka:


soxofaan:
It's bit unfortunate that we interpret an argument with name `filename` as a geojson object.

What was the use case again to support this feature?
Is it possible to introduce an alternative argument, e.g. `vector` for the usecase of passing geojson directly:
```
read_vector(filename="path/to/vector.json")  # -> Read file
read_vector(vector={"type" : "FeatureCollection" ....     # Use geojson directly
```

laxiwuka:
The crop calendar udp should work with file paths and with geojson so there is a read_vector in the process graph of this udp, but it the case of geojson the read_vector part can be ignored. As there is only one process graph, there should be an argument that accepts both cases, so maybe renaming the argument is an option, but I'm not sure if this wil break other things or not.

soxofaan:
I'm not sure if understand all the details, but another solution might be to to move the `read_vector` process out of the UDP and just declare a geojson parameter for the UDP. 
When calling the UDP you can still choose to pass a direct geojson object to this parameter, or to pass the result of a `read_vector` call. In pseudo-code:

```
// pass geojson directly
   "cropcal": {
       "process_id": "cropcal",
        "arguments": {
             "vector": {"type": "FeatureCollection" ...

// pass read_vector result
   "readvector": {
       "process_id": "read_vector",
       "arguments": {
            "filename": "path/to/vector.geojson"
       }
   }
   "cropcal": {
       "process_id": "cropcal",
        "arguments": {
             "vector": {"from_node": "readvector"}, ...
```

This some added benefits:
- read_vector is currently a custom VITO-only process, at some point an official process might be defined. That would work directly without having to change the UDP
- it allows to preprocess the geojson (e.g. with a UDF) before passing it to the UDP, again without having to change the UDP

jdries:
Moving read_vector out of the udp feels a bit like pushing too much complexity to the user.
read_vector is indeed not yet an official process, so should we perhaps start working on that, based on what we know now? 

Maybe indeed the proposal to not overload 'filename' but add a new argument, like 'featurecollection' instead is better? Not sure if both parameters can then be propagated to be optional UDP parameters?

soxofaan:
> Maybe indeed the proposal to not overload 'filename' but add a new argument, like 'featurecollection' instead is better? 

in that case, we probably should give it a more generic name like  `get_vector` or `get_geometry` (to align better with current terminology in the API).

> Not sure if both parameters can then be propagated to be optional UDP parameters?

I think that will work: you can give both UDP parameters a default value (null), so that `get_geometry` can detect which one is actually filled in by the user

m-mohr:
What is the exact use case of specifying a GeoJson in read_vector instead of just passing it to the corresponding process?

There's also load_uploaded_files, but I just realize it returns a raster-cube only, which seems wrong. 🤨

m-mohr:
> What is the exact use case of specifying a GeoJson in read_vector instead of just passing it to the corresponding process?

@soxofaan Could you answer this? I'm really wondering about the use case...

soxofaan:
I'm also not completely sure, but as far as I understood it is for a use case with a UDP where the end user wants to specify the geometry with a file path or a geojson object

jdries:
Yes, that's exactly it. For use cases (processing jobs) that require vector data, that data can be a file or online url, or simply a geojson. We want to build UDP's that support both, and as far as I know, openEO process graphs do not allow for complex logic to switch the selected process based on whether the input is a file or geojson. 
So allowing a process that can load or read vector data from some source seems like the most user friendly way.

m-mohr:
I see, thanks. It seems we would need a "get_type" process or so.

m-mohr:
Logged issue https://github.com/Open-EO/openeo-processes/issues/241

jdries:
Why do we need get_type? Solving the problem by adding more procedure logic in the process graphs is really not my preferred option in terms of simplicity.

m-mohr:
To support such use cases in general. With UDPs we will likely run into more of these issues in the future where you need to do different things depending on the parameter, think about distinguishing vector and raster cubes for example. I'm not saying you need this now, but I also don't see us adopting read_vector as it is added here. (I may be wrong.)

PR #67, opened 2021-04-30
Only use bbox from mask_polygon/aggregate_spatial when no explicit filter_bbox

soxofaan:
Follow up on this VITO backend feature that we implicitly apply `filter_bbox` with the bounding box of `aggregate_spatial`/`mask_polygon`, so that the end user can build a process graph with only `aggregate_spatial`/`mask_polygon` without worrying about correct `filter_bbox`.

Current implementation takes the bbox from first occurrence of `filter_bbox`/`aggregate_spatial`/`mask_polygon`. This could be confusing for user when they change order of operations (especially because associating an implicit `filter_bbox` with `mask_polygon` is a backend side optimization): `mask_polygon().filter_bbox()` vs `filter_bbox().mask_polygon()` might have different results or performance.

This PR should make things more predictable and stable:
- if there is an explicit `filter_bbox` (or spatial_extent in `load_collection`): use that, regardless of any other `aggregate_spatial`/`mask_polygon` processes
- if there is no explicit `filter_bbox`:  fallback on bbox of `aggregate_spatial`/`mask_polygon` (if any)

to do that, the bbox from  `aggregate_spatial`/`mask_polygon`  is stored as `_weak_spatial_extent` in the dry run which is only used as spatial extent when there is no explicit `spatial_extent`

### DIFF
--- a/openeo_driver/dry_run.py
+++ b/openeo_driver/dry_run.py
@@ -282,8 +282,8 @@ class DryRunDataTracer:
                         constraints["resample"] = {"target_crs": spatial_dim.crs, "resolution": resolutions}
 
             for op in [
-                "temporal_extent", "spatial_extent", "bands", "aggregate_spatial", "sar_backscatter",
-                "process_type", "custom_cloud_mask", "properties"
+                "temporal_extent", "spatial_extent", "_weak_spatial_extent", "bands", "aggregate_spatial",
+                "sar_backscatter", "process_type", "custom_cloud_mask", "properties"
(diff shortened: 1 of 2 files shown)


jdries (approved):
Agreed, this indeed enables full control over the bounding box, in combination with a polygon mask, and at the same time still allows the convenience of only specifying the polygon mask.

soxofaan:
sorry wrong button, I didn't want to re-request a review :)

PR #68, opened 2021-05-10
Remove old (POC) endpoints

soxofaan:
Remove old (POC) endpoints `/timeseries`, `/timeseries/point`, `/download`

@jdries   can we remove these old endpoints or are these still used somewhere internally?


soxofaan:
replaced by #93

PR #81, opened 2021-08-03
EP-3819 provide udf runtime to apply_tiles method

JeroenVerstraelen:
Provide runtime to apply_tiles method in order to support the PythonOffheap runtime.

### DIFF
--- a/openeo_driver/ProcessGraphDeserializer.py
+++ b/openeo_driver/ProcessGraphDeserializer.py
@@ -1148,7 +1148,7 @@ def _get_udf(args):
     udf = extract_arg(args, "udf")
     runtime = extract_arg(args, "runtime")
     # TODO allow registration of supported runtimes, so we can be more generic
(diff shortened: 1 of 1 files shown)


openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -1148,12 +1148,12 @@ def _get_udf(args):
     udf = extract_arg(args, "udf")
     runtime = extract_arg(args, "runtime")
     # TODO allow registration of supported runtimes, so we can be more generic
-    if runtime != "Python":
+    if not (runtime == "Python" or runtime == "PythonOffheap"):
```
jdries, line 1151: adjust runtime name, as in other package.
Maybe a backend needs to provide a list of runtime names to the python driver?

jdries (changes requested):
minor comment added

PR #85, opened 2021-09-20
EP-4012: moved creo catalog from openeo-geopyspark-driver and added i…

laxiwuka:
…mplementation for /validation

### DIFF
--- a/openeo_driver/dry_run.py
+++ b/openeo_driver/dry_run.py
@@ -34,8 +34,7 @@ These source constraints can then be fetched from the EvalEnv at `load_collectio
 """
 import logging
 from enum import Enum
-from datetime import datetime
-from pathlib import Path
+from datetime import datetime, time
 from typing import List, Union, Tuple
 
 import shapely.geometry.base
@@ -43,6 +42,7 @@ import shapely.geometry.base
 from openeo.metadata import CollectionMetadata
 from openeo_driver import filter_properties
 from openeo_driver.catalogs.creo import CatalogConstants, CatalogClient
+from openeo_driver.catalogs.oscars import OscarsCatalogClient
 from openeo_driver.datacube import DriverDataCube
 from openeo_driver.datastructs import SarBackscatterArgs, ResolutionMergeArgs
 from openeo_driver.delayed_vector import DelayedVector
@@ -121,12 +121,12 @@ class DataSource(DataTraceBase):
         return self._process
 
     @classmethod
-    def load_collection(cls, collection_id, properties={}) -> 'DataSource':
+    def load_collection(cls, collection_id, properties={}, source_info={}) -> 'DataSource':
         """Factory for a `load_collection` DataSource."""
         exact_property_matches = {property_name: filter_properties.extract_literal_match(condition)
                                   for property_name, condition in properties.items()}
 
-        return cls(process="load_collection", arguments=(collection_id, exact_property_matches))
+        return cls(process="load_collection", arguments=(collection_id, exact_property_matches, source_info))
 
     @classmethod
     def load_disk_data(cls, glob_pattern: str, format: str, options: dict) -> 'DataSource':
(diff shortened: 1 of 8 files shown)


openeo_driver/dry_run.py
```diff
@@ -532,6 +535,27 @@ def atmospheric_correction(self, method: str = None, *args) -> 'DriverDataCube':
     def mask_scl_dilation(self) -> 'DriverDataCube':
         return self._process("custom_cloud_mask", arguments={"method": "mask_scl_dilation"})
 
+    def get_missing_products(self):
+        missing_products = []
+
+        if self.metadata.get("id") == "SENTINEL2_L2A":
+            opensearch_endpoint = self.metadata.get("_vito", "data_source", "opensearch_endpoint")
```
soxofaan, line 542: I'm afraid that `self.metadata` will not always be usable/available.
From the dry run you should only trust the data from `DryRunDataTracer`

openeo_driver/dry_run.py
```diff
@@ -532,6 +535,27 @@ def atmospheric_correction(self, method: str = None, *args) -> 'DriverDataCube':
...
+            if opensearch_endpoint and "creo" in opensearch_endpoint:
+                mission = CatalogConstants.missionSentinel2
+                level = CatalogConstants.level2A
+                catalog = CatalogClient(mission, level)
+                source_constraints = self._data_tracer.get_source_constraints()
+                temporal_extent = source_constraints[0][1]['temporal_extent']
```
soxofaan, line 548: you should not only look at the first element of `source_constraints`, but all

openeo_driver/dry_run.py
```diff
@@ -532,6 +535,27 @@ def atmospheric_correction(self, method: str = None, *args) -> 'DriverDataCube':
     def mask_scl_dilation(self) -> 'DriverDataCube':
         return self._process("custom_cloud_mask", arguments={"method": "mask_scl_dilation"})
 
+    def get_missing_products(self):
+        missing_products = []
+
+        if self.metadata.get("id") == "SENTINEL2_L2A":
```
soxofaan, line 541: can we avoid this hardcoded collection id `SENTINEL2_L2A`?

openeo_driver/dry_run.py
```diff
@@ -532,6 +535,27 @@ def atmospheric_correction(self, method: str = None, *args) -> 'DriverDataCube':
     def mask_scl_dilation(self) -> 'DriverDataCube':
         return self._process("custom_cloud_mask", arguments={"method": "mask_scl_dilation"})
 
+    def get_missing_products(self):
```
soxofaan, line 538: I wonder if it is possible to avoid defining this method on `DryRunDataCube`, 
not sure what alternatives there are

soxofaan (commented):
sorry for delay, I just found some time for a quick review

openeo_driver/dry_run.py
laxiwuka: I now pass the `_vito data_source` info to `DataSource.load_collection` so I can get that info from `source_constraints`.

openeo_driver/dry_run.py
laxiwuka: I now look at all `load_collection` elements in the `source_constraints`.

openeo_driver/dry_run.py
laxiwuka: Not sure how I can avoid this check.

openeo_driver/dry_run.py
laxiwuka: As I don't need the metadata of the `DryRunDataCube` anymore, I moved this method to `DryRunDataTracer`. Is this a better place for it?

soxofaan:
closing this issue.
Content of PR has been split in a part in Open-EO/openeo-python-driver and a part in Open-EO/openeo-geopyspark-driver in other branches (now merged in master)

PR #87, opened 2021-09-30
EP-3993: register root path in flask app

laxiwuka:


openeo_driver/views.py
```diff
@@ -208,6 +208,7 @@ def well_known_openeo():
             blueprint=bp, backend_implementation=backend_implementation, api_endpoint=api_reg, auth_handler=auth
         )
 
+    app.register_blueprint(bp, url_prefix='/')
```
soxofaan, line 211: I would not do that,
this registers all openEO endpoints directly under `/`, which would conflict with adding non-standard endpoints/pages or adding support for future openEO API versions.

at the moment I would just add temp (302) forwarding of `/` to `/openeo/1.0`

PR #97, opened 2021-11-18
EP-4067/EP-4068: add drop_dimension and dimension_labels

laxiwuka:


openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -607,6 +607,18 @@ def add_dimension(args: dict, env: EvalEnv) -> DriverDataCube:
     )
 
 
+@process_registry_100.add_function
+def drop_dimension(args: dict, env: EvalEnv) -> DriverDataCube:
+    data_cube = extract_arg(args, 'data')
+    return data_cube.drop_dimension(dimension=extract_arg(args, 'dimension'))
```
soxofaan, line 613: per https://processes.openeo.org/#drop_dimension, `drop_argument`'s argument name is `name`, not `dimension`

openeo_driver/datacube.py
```diff
@@ -51,6 +51,12 @@ def reduce_dimension(self, reducer, dimension: str, env: EvalEnv) -> 'DriverData
     def add_dimension(self, name: str, label, type: str = "other") -> 'DriverDataCube':
         self._not_implemented()
 
+    def drop_dimension(self, dimension: str = "other") -> 'DriverDataCube':
```
soxofaan, line 54: as noted above: per spec, the argument name should be `name`, not `dimension`

openeo_driver/dry_run.py
```diff
@@ -485,6 +485,12 @@ def reduce_dimension(self, reducer, dimension: str, env: EvalEnv) -> 'DryRunData
     def add_dimension(self, name: str, label, type: str = "other") -> 'DryRunDataCube':
         return self._process_metadata(self.metadata.add_dimension(name=name, label=label, type=type))
 
+    def drop_dimension(self, dimension: str = "other") -> 'DryRunDataCube':
```
soxofaan, line 488: as noted above: per spec, the argument name should be name, not dimension

openeo_driver/ProcessGraphDeserializer.py
soxofaan: (also see https://github.com/Open-EO/openeo-processes/issues/305)

PR #113, opened 2022-04-14
Return stac metadata for save ml model process (geopyspark driver issue 139)

JeroenVerstraelen:


tests/test_views.py
```diff
@@ -1320,7 +1349,7 @@ def test_get_job_results_signed_with_expiration_110(self, api110, flask_app):
             assert resp.assert_status_code(200).json == {
                 'type': 'Collection',
                 'stac_version': '0.9.0',
-                'stac_extensions': ['eo', 'file'],
+                'stac_extensions': ['eo', 'file', 'https://stac-extensions.github.io/ml-model/v1.0.0/schema.json'],
```
soxofaan, line 1352: I'm confused, why wasnt this extra stac_extension necessary for previous test?

tests/test_views.py
```diff
@@ -1350,6 +1379,11 @@ def test_get_job_results_signed_with_expiration_110(self, api110, flask_app):
                         'type': 'application/geo+json'
                     }
                 ],
+                'summaries': {
+                    'ml-model:architecture': 'random-forest',
```
soxofaan, line 1383: I think the summary values should be list of strings instead of just strings

tests/test_views.py
```diff
@@ -1065,7 +1068,12 @@ def test_get_job_results_100(self, api100):
                             'center_wavelength': 1.23
                         }],
                         'file:nodata':[123]
-                    }
+                    },
+                    'ml_model_metadata.json': {
+                        'href': 'http://oeo.net/openeo/1.0.0/jobs/07024ee9-7847-4b8a-b260-6c879a2b3cdc/results/ml_model_metadata.json',
+                        'rel': 'item',
```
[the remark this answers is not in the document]
JeroenVerstraelen: This follows the ML model extension specification:
https://github.com/stac-extensions/ml-model/blob/main/examples/cyclone-wind-speed/collection.json

However I just noticed that their example is likely wrong and does not follow their schema:
https://github.com/stac-extensions/ml-model/blob/main/json-schema/schema.json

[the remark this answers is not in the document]
soxofaan: I see you already changed it, but you originally used `'rel': 'item', ` in under `"assets"` , while "rel" is a typical field in link objects under "links"

openeo_driver/views.py
```diff
@@ -847,10 +849,11 @@ def list_job_results(job_id, user: User):
             })
 
             assets = {filename: _asset_object(job_id, user.user_id, filename, asset_metadata)
-                      for filename, asset_metadata in results.items()}
+                      for filename, asset_metadata in results.items() if asset_metadata.get('asset', False)}
```
soxofaan, line 852: Do I understand correctly that the "asset" field is False by default now, while previously everything was assumed to be an asset?  Isn't it risky to change the default behavior here?

openeo_driver/views.py
```diff
@@ -881,6 +893,17 @@ def list_job_results(job_id, user: User):
...
                     "assets": assets
                 })
+                if ml_model_metadata is not None:
+                    result["stac_extensions"].extend(ml_model_metadata.get("stac_extensions", []))
+                    if "summaries" not in result.keys():
+                        result["summaries"] = {}
+                    if "properties" in ml_model_metadata.keys():
+                        ml_model_properties = ml_model_metadata["properties"]
+                        result["summaries"].update({
+                            "ml-model:learning_approach": ml_model_properties.get("ml-model:learning_approach", []),
```
soxofaan, line 903: these summaries should be a list, while `ml_model_properties.get("ml-model:learning_approach")`  returns a string I assume

openeo_driver/views.py
```diff
@@ -990,25 +1013,58 @@ def _download_job_asset_stac_item(job_id: str, stac_item_filename: str, user: Us
         return resp
 
     def _asset_object(job_id, user_id, filename: str, asset_metadata: dict) -> dict:
+        result_dict = dict_no_none(**{
```
soxofaan, line 1016: no need for `dict_no_none(**{`, just do `dict_no_none({`

openeo_driver/views.py
```diff
@@ -990,25 +1013,58 @@ def _download_job_asset_stac_item(job_id: str, stac_item_filename: str, user: Us
...
+                                                                   filename=model_path, _external=True)
+        except KeyError:
+            pass
+
+        stac_item = {
+            "type": "Feature",
+            "stac_version": "0.9.0",
+            "collection": job_id,
+        }
+        ml_model_metadata.pop('ml_model_metadata', None)
```
soxofaan, line 1052: I think it's cleaner and more self-describing to explicitly copy the fields you expect to a new dict, instead of removing (`pop`) one particular field from an exisiting dict and blindly assuming the other fields are ok.

openeo_driver/dummy/dummy_backend.py
```diff
@@ -467,12 +467,80 @@ def get_results(self, job_id: str, user_id: str) -> Dict[str, dict]:
...
                 "roles": ["data"],
                 "bands": [Band(name="NDVI", common_name="NDVI", wavelength_um=1.23)],
                 "nodata": 123,
                 "instruments": "MSI"
+            },
+            "randomforest.model": {
+                "asset": True,
+                "href": str(job_id / Path("randomforest.model")),
+            },
+            "ml_model_metadata.json": {
```
soxofaan, line 482: This should be `DriverMlModel.METADATA_FILE_NAME` I think

soxofaan (commented):
some more notes

soxofaan:
(FYI this is dependency for https://github.com/Open-EO/openeo-geopyspark-driver/issues/139)

openeo_driver/views.py
```diff
@@ -990,6 +1003,12 @@ def _download_job_asset_stac_item(job_id: str, stac_item_filename: str, user: Us
         return resp
 
     def _asset_object(job_id, user_id, filename: str, asset_metadata: dict) -> dict:
+        if filename == "ml_model_metadata.json":
+            return dict_no_none(**{
```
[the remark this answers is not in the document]
JeroenVerstraelen: Solved.

tests/test_views.py
```diff
@@ -1209,7 +1225,12 @@ def test_get_job_results_signed_100(self, api100, flask_app):
                             'center_wavelength': 1.23
                         }],
                         'file:nodata': [123]
-                    }
+                    },
+                    'ml_model_metadata.json': {
+                        'href': 'http://oeo.net/openeo/1.0.0/jobs/07024ee9-7847-4b8a-b260-6c879a2b3cdc/results/ml_model_metadata.json',
```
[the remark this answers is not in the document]
JeroenVerstraelen: I believe this comment was referencing _test_get_job_results_040_.  But I don't see why ml_model_metadata.json should use a signed url here. There is no app_config in this test, like for example in _test_get_job_results_signed_040_:
`app_config = {'SIGNED_URL': 'TRUE', 'SIGNED_URL_SECRET': '123&@#'}`

tests/test_views.py
JeroenVerstraelen: The ml-model stac_extension is only added if the api version is atleast 1.1.0 and the batch job dir contains a ml_model_metadata.json file.

- The 0.4.0 version only has a 'links' section so the ml_model and ml_model_metadata just get added to those links.
- In the 1.0.0 version only the ml_model is added but the ml_model_metadata is not added to the list of links + the job metadata does not add the ml-model stac extension (with summaries, etc.).
- The 1.1.0 version does everything correctly, with a ml-model stac extension, ml_model in the assets, and ml_model_metadata in the links.

Perhaps I should add the ml-model stac extensions to the 1.0.0 version as well?

openeo_driver/views.py
```diff
@@ -1003,12 +1022,38 @@ def _asset_object(job_id, user_id, filename: str, asset_metadata: dict) -> dict:
...
+        results = backend_implementation.batch_jobs.get_results(job_id, user.user_id)
+        ml_model_metadata: dict = results.get(file_name, None)
+        if ml_model_metadata is None:
+            return jsonify({})
+        try:
+            model_path = pathlib.Path(ml_model_metadata["assets"]["model"]["href"]).name
+            ml_model_metadata["assets"]["model"]["href"] = url_for('.download_job_result', job_id=job_id,
+                                                                   filename=model_path, _external=True)
+        except KeyError:
+            pass
```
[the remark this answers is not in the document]
JeroenVerstraelen: Because the href of a model should only be changed if it exists.

I could have used:
`if ml_model_metadata.get("assets", {}).get("model", {}).get("href", None) is not None:
`or
```
def deep_get(dictionary, *keys):
    return functools.reduce(lambda d, key: d.get(key) if d else None, keys, dictionary)
if deep_get(ml_model_metadata, "assets", "model", "href") is not None:
```

But a try except should be the most pythonic solution:
https://stackoverflow.com/questions/43491287/elegant-way-to-check-if-a-nested-key-exists-in-a-dict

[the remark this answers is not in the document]
soxofaan: FYI: we already have `deep_get` in `openeo.util` and it's already imported in views.py

I understand the proposed solution of that stackoverlfow post, but the big difference is that the indexing there is the only thing happening, and it is happening on the right hand side. Then a try-catch is an easy solution indeed.
But here there is a `url_for` function call and deeply nested assignment (indexing on left hand side). Doing try-catch here looks pretty confusing. Is the exception coming from `url_for`?  Is there a intermediate level missing? Is the deepest level ("href") missing?

openeo_driver/views.py
JeroenVerstraelen: Solved.

openeo_driver/views.py
JeroenVerstraelen: Solved.

openeo_driver/views.py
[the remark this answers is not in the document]
JeroenVerstraelen: Thanks, I did not see that we already had a deep_get implementation. I changed it so it uses that function.

PR #117, opened 2022-04-20
add support to load ml model using (signed) url

JeroenVerstraelen:


openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -603,8 +603,18 @@ def save_ml_model(args: dict, env: EvalEnv) -> MlModelResult:
...
 @process_registry_100.add_function(spec=read_spec("openeo-processes/experimental/load_ml_model.json"))
 def load_ml_model(args: dict, env: EvalEnv) -> DriverMlModel:
-    job_id = extract_arg(args, "id")
-    return env.backend_implementation.load_ml_model(job_id)
+    from urllib.parse import urlparse
+    model_id = extract_arg(args, "id")
+    if model_id.startswith('http'):
+        now = datetime.datetime.now()
+        now_hourly_truncated = now - datetime.timedelta(minutes=now.minute, seconds=now.second, microseconds=now.microsecond)
+        hourly_id = hash(model_id + str(now_hourly_truncated))
```
jdries, line 611: This will download the model to a shared directory, that doesn't exist on CreoDIAS.
Would it be possible to do the download in each executor separately?

JeroenVerstraelen: This was made to be analogous to the implementation in DelayedVector. Is it okay if I create a new directory under the /tmp/ directory of each executor? The final file would then be e.g. /tmp/download_{hourly_id}/randomforest.model

jdries: For temporary directories in Python, you can use https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp
In Spark, you can also write to the current working directory, which you can assume to be cleaned up.

soxofaan:
I'm not sure it is necessary at this moment that all this logic has to be in the python driver. I think you just have to properly support HTTP model references in the `load_ml_model` implementation in  geopyspark driver.

Also, I think this implementation does the download twice because of the dry run phase, unless I'm missing something

soxofaan:
Another thing: in case of geopyspark driver: the download will happen in the spark driver, while you need the file in the executors and I'm not sure that you can assume the `/tmp`  trees are shared between the driver and the executors. I think that is what Jeroen referred to in 

> Would it be possible to do the download in each executor separately?

openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -603,8 +603,15 @@ def save_ml_model(args: dict, env: EvalEnv) -> MlModelResult:
...
 def load_ml_model(args: dict, env: EvalEnv) -> DriverMlModel:
-    job_id = extract_arg(args, "id")
-    return env.backend_implementation.load_ml_model(job_id)
+    from urllib.parse import urlparse
+    model_id = extract_arg(args, "id")
+    if model_id.startswith('http'):
+        download_directory = tempfile.mkdtemp()
+        dest_path = Path(download_directory + "/" + str(urlparse(model_id).path.split("/")[-1]))
+        with open(dest_path, 'wb') as f:
+            f.write(requests.get(model_id).content)
```
soxofaan, line 612: if I understand the `load_ml_model` correctly (https://github.com/Open-EO/openeo-processes/blob/draft/proposals/load_ml_model.json)  the URL is the URL of the STAC Item, and not the URL of the model file itself. So you have to parse the STAC too to and extract the actual model URL

JeroenVerstraelen: This might require a discussion with @m-mohr, because we have a link to our ml-model itself in the assets section of our batch job metadata. (E.g. randomforest.model) It seems more transparent for the user if they can just use the link to the ml-model itself, rather than a link to its STAC metadata.

jdries: can we somehow get a working version today still?

m-mohr: This was indeed meant to use the STAC Item (i.e. your batch job metadata). The thought behind it is that additional metadata may be required to load the ml model properly. Not sure whether that is required in your case though, but it might be required to load it from another back-end.

JeroenVerstraelen:
Closing this PR because the implementation can be done in the geopyspark-driver.

PR #157, opened 2022-11-29
Issue #232 k8s manage job results directly on object storage

JohanKJSchreurs:
# Implement Issue #232, k8s manage job results directly on object storage

These are the modifications on the Python driver side to implement issue [Open-EO/openeo-geopyspark-driver/ #232](https://github.com/Open-EO/openeo-geopyspark-driver/issues/232)

This issue requires changes in both the openeo-python-driver and the openeo-geopyspark-driver.

## Changed

When the path to the result file is a URL on S3 (i.e. it starts with "s3://" ) then the file with be retrieved directly from S3 and wrapped in a streaming Flask response. 

## N.B. New dependencies

 Added boto3 to retrieve files from S3 and moto to have a mock for S3 in the unit tests.


openeo_driver/dummy/dummy_backend.py
```diff
@@ -605,16 +605,16 @@ def _update_status(cls, job_id: str, user_id: str, status: str):
     def start_job(self, job_id: str, user: User):
         self._update_status(job_id=job_id, user_id=user.user_id, status="running")
 
-    def _output_root(self) -> Path:
-        return Path("/data/jobs")
+    def _output_root(self) -> str:
+        return "/data/jobs"
```
jdries, line 609: Isn't this less cross-platform then the previous Path-based version?

soxofaan: Indeed, I would also prefer a smarter `Path` over a dumb `str` by default, 
which you then downgrade where necessary with `str(path)`

openeo_driver/views.py
```diff
@@ -2,11 +2,14 @@
...
 import json
 import logging
+import os
 import pathlib
 import re
+import requests
 from collections import namedtuple, defaultdict
 from typing import Callable, Tuple, List, Optional
 
+import boto3
```
soxofaan, line 12: as discussed: can we avoid this s3-specific import from views.py?

openeo_driver/dummy/dummy_backend.py
JohanKJSchreurs: Indeed it is.  The reason I made this change was to support a s3:// URL, but maybe there is a better way to support both a path and a URL. 
First I tried to create a Path using the URL, but that doesn't work ( s:// becomes s:/ )

tests/test_views.py
```diff
@@ -73,6 +74,42 @@ def hex(self):
...
+
+TEST_AWS_REGION_NAME = 'eu-central-1'
+import os
+import boto3
+from moto import mock_s3
+
+@pytest.fixture(scope='function')
+def aws_credentials():
+    """Mocked AWS Credentials and related environment variables for moto/boto3."""
+    os.environ['AWS_ACCESS_KEY_ID'] = 'testing'
```
soxofaan, line 86: This fixture does not reset `os.environ` on exit,  
see https://github.com/Open-EO/openeo-python-client/blob/03fdfcb6ee5be2c8c08bec89e77ed1913205a642/tests/test_config.py#L93  for how to do this  with `mock.patch.dict`

openeo_driver/dummy/dummy_backend.py
soxofaan: ah ok, `Path` indeed does not support s3 scheme probably. String is fine then

tests/test_views.py
```diff
@@ -73,6 +74,42 @@ def hex(self):
     return mock.patch("uuid.uuid4", new=UUIDMock)
 
 
+
+TEST_AWS_REGION_NAME = 'eu-central-1'
+import os
+import boto3
+from moto import mock_s3
```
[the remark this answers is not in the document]
JohanKJSchreurs: Ah, I missed that one.

Related to that:
Would it be a good idea to start using [isort](https://pycqa.github.io/isort/) to sort our import in a logical order?
That's one of those things that normally makes things more readable and can be done automatically and, not something you want to waste your brains energy for. 

However, if we would automatically sort imports we might need to check if there are some issues with import order and circular dependencies. Because I do remember some files (maybe in another openeo codebase) where local imports inside methods were needed to avoid circular import problems.

tests/test_views.py
JohanKJSchreurs: True, that fixture should restore the environment variables after use.

FYI: pytest's monkeypatch also has a handy way to do just that. Just another option.  
https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html

tests/test_views.py
```diff
@@ -73,6 +74,42 @@ def hex(self):
...
+
+
+@pytest.fixture(scope='function')
+def mock_s3_resource(aws_credentials):
+    with mock_s3():
+        yield boto3.resource("s3", region_name=TEST_AWS_REGION_NAME)
+
+
+@pytest.fixture(scope='function')
+def mock_s3_client(aws_credentials):
```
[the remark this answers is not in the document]
JohanKJSchreurs: Indeed, not used. Looks like a leftover. We don't need it so I'll remove it.

tests/test_views.py
soxofaan: ah interesting, I didn't know monkeypatch from pytest yet, 
looks indeed like an alternative solution here

tests/test_views.py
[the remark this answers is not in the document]
soxofaan: When touching code I often use the "optimize imports" feature of PyCharm (https://www.jetbrains.com/pycharm/guide/tips/optimize-imports/,  `ctrl-alt-o` shortcut for me)  unless it ruins the signal/noise ratio of my commit. It is very close (a bit less strict) to what isort does.

I recently also started using [darker](https://github.com/akaihola/darker)   to black-ify (and sometimes isort-ify) the code regions I touched. (I've set up a keyboard shortcut for it in Pycharm  as documented at https://github.com/akaihola/darker#pycharmintellij-idea .) Very handy to automate/standardize code style and consistency.

I would certainly be in favor to push this more team-wide, but I'm not sure what the best approach is to promote/enforce this.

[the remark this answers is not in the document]
JohanKJSchreurs: Currently I'm using VSCode but there are probably similar solutions there to optimize import Or I just use isort from the CLI. For the time being VSCode gives me better integration with Docker and WSL. I just haven't spent the time yet to figure all of that out in PyCharm.

If we were to standardize such tools as darker, yapf, flake8, isort etc, one way is to make it simpler to run them is [pre-commit (the tool) ](https://pre-commit.com/)

You would configure a set of tools to run in a config file, and pre-commit can then run all of them in one go. This is typically set up to run on git's pre-commit hook but I have previously used it without the hook just as a command line tool you run when you want to.

Pre-commit is completely independent of what IDE you use, and because it is just automation you could set it up to run the same checks as on your CI/CD so you flush out more things early on.

Added benefit, there are also hooks to also integrate security tools such as  [bandit](https://bandit.readthedocs.io/en/latest/index.html), to detect that you are using vulnerable versions of dependencies that need to be upgraded.

openeo_driver/views.py
```diff
@@ -722,6 +728,18 @@ def _properties_from_job_info(job_info: BatchJobMetadata) -> dict:
     return properties
 
 
+def s3_client() -> boto3.client:
+    """Create an S3 client to access object storage on Swift."""
+    aws_access_key_id = os.environ.get("AWS_ACCESS_KEY_ID")
+    aws_secret_access_key=os.environ.get("AWS_SECRET_ACCESS_KEY")
+    swift_url = os.environ.get("SWIFT_URL")
```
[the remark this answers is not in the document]
JohanKJSchreurs: Done.

openeo_driver/views.py
JohanKJSchreurs: I have moved the boto3 import inside the function that creates the s3 client, `_s3_client()`

And in setup.py the requirement for the boto3 dependency has been moved to a new `extra_require` item, named "s3".
For dev / unit testing we still need both boto3 and moto, so these are both present in the test_require. 

I considered moving the S3 related functions to a separate python module, but in the end there was only one function left that would live there, namely `_s3_client()`  Did make it an internal / private function though. Better to be conservative about what is available as public.

tests/test_views.py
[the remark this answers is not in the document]
JohanKJSchreurs: Is now removed.

PR #173, opened 2023-03-24
281 account sh pu for result calls

bossie:
Implements openeo-python-driver part for https://github.com/Open-EO/openeo-geopyspark-driver/issues/281.


openeo_driver/views.py
```diff
@@ -606,28 +606,44 @@ def result(user: User):
...
+            else:
+                if not isinstance(result, SaveResult):
+                    # Implicit save result (using default/best effort format and options)
+                    result = to_save_result(data=result)
+                response = result.create_flask_response()
+
+            # not all costs are accounted for so don't expose in "OpenEO-Costs" yet
+            request_costs(success=True)
+        except Exception:
+            request_costs(success=False)
```
soxofaan, line 643: `request_costs`  is defined to return `Optional[float]`, but that return value is not being used here, so I'm not sure what the purpose of this is?

openeo_driver/backend.py
```diff
@@ -705,5 +707,11 @@ def changelog(self) -> Union[str, Path]:
...
 
-    def after_request(self):
+    def after_request(self, request_id: str):
         pass
+
+    def set_request_costs_getter(self, get_request_costs: Callable[[str, str, bool], Optional[float]]):
+        self._get_request_costs = get_request_costs
+
+    def request_costs(self, user_id: str, request_id: str, success: bool) -> Optional[float]:
+        return self._get_request_costs(user_id, request_id, success)
```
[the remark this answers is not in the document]
bossie: Simply overriding will make every `GeoPySparkBackendImplementation` use the ETL API whereas we want to enable it for Terrascope/Creo en disable it for other environments (~ https://github.com/Open-EO/openeo-geopyspark-driver/pull/331).

I guess passing a parameter to the `GeoPySparkBackendImplementation` constructor would work as well or do you have something else in mind?

openeo_driver/views.py
bossie: > `request_costs` is defined to return `Optional[float]`, but that return value is not being used here, so I'm not sure what the purpose of this is?

It's in the comment: initially I just passed the SHub PUs to the ETL API and returned its costs response in the `OpenEO-Costs` header but then realized that there's also CPU, memory etc to consider so the value in this header would not (yet) be correct and I removed it. I don't see an immediate reason to change the method to populate it (`request_costs`) though.

bossie:
For the complete picture, consider these PRs as well: https://github.com/Open-EO/openeo-geopyspark-driver/pull/379 and https://github.com/Open-EO/openeo-geotrellis-extensions/pull/141.

PR #174, opened 2023-03-30
Fix to prevent: "log collection failed: circuit breaking exception"

JohanKJSchreurs:
# Fix to prevent: "log collection failed: circuit breaking exception"

Part of implementation to fix issue #170

##  Add log_level parameter to get_log_entries.

Setting log_level should help to limit the amount of memory used in Elasticsearch while retrieving logs.
The cause of the circuit_breaker_exception is that amount of memory being used to query the logs, is getting too close to the maximum limit allowed.

This circuit breaker is apparently a mechanism to prevent that the whole Elasticsearch process runs out of memory which would make it crash. Therefore the request is cut off before it really exhausts memory.

There seem to be either too many log records, or there may be fields that contain a large amount of data (long messages perhaps).

## Multiple code bases involved

This has to be implemented throughout three repositories each with there respective PR.
This PR is the main PR (because the issue was logged on the openeo-python-driver)

- https://github.com/Open-EO/openeo-python-driver : 
    * The main PR.
    * The Python driver needs to link through log_level parameter to backend implementation in GeoPySpark driver.
- https://github.com/Open-EO/openeo-geopyspark-driver :
    *  Where logs are actually retrieved from Elasticsearch.
    *  PR: https://github.com/Open-EO/openeo-geopyspark-driver/pull/382
- https://github.com/Open-EO/openeo-python-client  
    * needs to be use log_level when retrieving logs via the openeo-python-driver
    * PR: https://github.com/Open-EO/openeo-python-client/pull/408


openeo_driver/backend.py
```diff
@@ -439,7 +439,13 @@ def get_results(self, job_id: str, user_id: str) -> Dict[str, dict]:
         # TODO: eliminate this method in favor of `get_result_assets`
         raise NotImplementedError
 
-    def get_log_entries(self, job_id: str, user_id: str, offset: Optional[str] = None) -> Iterable[dict]:
+    def get_log_entries(
+        self,
+        job_id: str,
+        user_id: str,
+        offset: Optional[str] = None,
+        log_level: Optional[str] = None,
```
soxofaan, line 447: FYI: request parameter will be `level` iso `log_level` (https://github.com/Open-EO/openeo-api/pull/488) 
it's probably better to stick to that convention everywhere

openeo_driver/views.py
```diff
@@ -1258,9 +1258,12 @@ def download_job_result_signed(job_id, user_base64, secure_key, filename):
     @blueprint.route('/jobs/<job_id>/logs', methods=['GET'])
     @auth_handler.requires_bearer_auth
     def get_job_logs(job_id, user: User):
-        offset = request.args.get('offset')
+        offset = request.args.get("offset")
+        log_level = request.args.get("log_level")
```
soxofaan, line 1262: as noted above : `log_level` -> `level`

soxofaan (commented):
apart from `log_level` -> `level`, this looks ready for merging

openeo_driver/backend.py
JohanKJSchreurs: Done.

openeo_driver/dummy/dummy_backend.py
```diff
@@ -731,11 +732,21 @@ def get_result_assets(self, job_id: str, user_id: str) -> Dict[str, dict]:
...
+        user_id: str,
+        offset: Optional[str] = None,
+        level: Optional[str] = None,
+    ) -> Iterable[dict]:
         self._get_job_info(job_id=job_id, user_id=user_id)
         default_logs = [{"id": "1", "level": "info", "message": "hello world"}]
         for log in self._custom_job_logs.get(job_id, default_logs):
             if isinstance(log, dict):
+                actual_level = normalize_log_level(log.get("log_level"))
+                requested_level = normalize_log_level(level)
```
soxofaan, line 747: `requested_level` can be moved outside of for loop I think

soxofaan (commented):
just minor note, looks ready for merge

I would also bump the minor version level in `_version.py` to mark this as an API change (which can be depended on in geopyspark driver)

openeo_driver/dummy/dummy_backend.py
JohanKJSchreurs: Indeed, no need to repeat that function call.  Done.
Minor version bumped as well.

Issue #176, opened 2023-04-11
Fix pandas 2.0 compatibility


JohanKJSchreurs:
There are some new errors with Pandas 2.0, while we don't have those errors with Pandas 1.5.

The pandas dependency is currently not pinned to a version.
We could start to pin it, but at present that would need to be pinned to version 1.5.x to make the tests pass. That means we will still have a problem later on when we upgrade to Pandas 2.0.

So it would be best to fix the issues and make the Python driver compatible with Pandas 2.0.x.

## Example error logs from pytest

Error 1:

Can be traced back to a breaking change in the Pandas API: 

See also: https://pandas.pydata.org/docs/whatsnew/v2.0.0.html#backwards-incompatible-api-changes
> Removed the closed argument in [date_range()](https://pandas.pydata.org/docs/reference/api/pandas.date_range.html#pandas.date_range) and [bdate_range()](https://pandas.pydata.org/docs/reference/api/pandas.bdate_range.html#pandas.bdate_range) in favor of inclusive argument ([GH40245](https://github.com/pandas-dev/pandas/issues/40245))

```
=================================== FAILURES ===================================
___________________________ test_period_to_intervals ___________________________

...
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

start = Timestamp('2021-06-01 00:00:00'), end = Timestamp('2021-06-24 00:00:00')
periods = None, freq = 'W', tz = None, normalize = False, name = None
inclusive = 'both', unit = None, kwargs = {'closed': 'left'}

    def date_range(
        start=None,
        end=None,
        periods=None,
        freq=None,
        tz=None,
        normalize: bool = False,
        name: Hashable = None,
        inclusive: IntervalClosedType = "both",
        *,
        unit: str | None = None,
        **kwargs,
    ) -> DatetimeIndex:
        """
        Return a fixed frequency DatetimeIndex.
     
      # Removed part of this docstring for brevity
    
        >>> pd.date_range(start="2017-01-01", periods=10, freq="100AS", unit="s")
        DatetimeIndex(['2017-01-01', '2117-01-01', '2217-01-01', '2317-01-01',
                       '2417-01-01', '2517-01-01', '2617-01-01', '2717-01-01',
                       '2817-01-01', '2917-01-01'],
                      dtype='datetime64[s]', freq='100AS-JAN')
[...]

PR #177, opened 2023-04-11
Fix: some unit tests failed when using Pandas 2.0

JohanKJSchreurs:
Fixes issue #176.

Some fixes to make the code compatible with pandas 2.0.

Tests now pass with both pandas v1.5 and v2.0.

For now, I have not yet pinned the pandas version, because we might want to discuss first if we want to leave it unpinned, or what version constraint we should apply.   (I suppose we would either pin it to `pandas>=2.0` or to `pandas<=2.0` )

I don't know if we have users who may be stuck on pandas 1.5.

### DIFF
--- a/openeo_driver/ProcessGraphDeserializer.py
+++ b/openeo_driver/ProcessGraphDeserializer.py
@@ -1081,17 +1081,17 @@ def _period_to_intervals(start, end, period) -> List[Tuple[pd.Timestamp, pd.Time
     # TODO: "hour" support?
     if "day" == period:
         offset = timedelta(days=1)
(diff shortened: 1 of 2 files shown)


soxofaan:
As long as there is no concrete need to add a constraint at the level of openeo-python-driver library, I would leave things as-is.
If we need constraints  for the Terrascope/creodias deploys, we can do that in their respective requirements.txt files (e.g. in openeo-deploy project)

soxofaan:
I've set the pandas constraint to `>=1.3.0` for now (1.3.5 works on my machine), to at least have some constraint.

pandas 2 is relatively young at the moment, so maybe better wait it out a bit before deciding to jump ship completely to `>=2.0`

PR #181, opened 2023-04-17
Add STAC projection metadata fields in job results metadata.

JohanKJSchreurs:
Previous PRs have added STAC projection metadata to the metadata file of the jobs results.
However the API selectively copies fields from that metadata on file to a new dictionary that is the one that the API returns,
for  `GET /jobs/<job_id>/results/` request

This PR adds those fields to that result.

Part of the fix for issue: https://github.com/Open-EO/openeo-geotrellis-extensions/issues/72

### DIFF
--- a/openeo_driver/views.py
+++ b/openeo_driver/views.py
@@ -1238,11 +1238,29 @@ def register_views_batch_jobs(
         bands = asset_metadata.get("bands")
         nodata = asset_metadata.get("nodata")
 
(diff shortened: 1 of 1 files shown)


jdries:
Merged it as I'd like to have this metadata. We can perhaps improve this selective copy mechanism and simply copy all metadata that is not handled by the selective approach.

PR #184, opened 2023-04-25
Pass through projection metadata at STAC item level

JohanKJSchreurs:
Is part of issue: https://github.com/Open-EO/openeo-geotrellis-extensions/issues/72

Needed by PR https://github.com/Open-EO/openeo-geopyspark-driver/pull/405 because the GeoPySpark driver needs to pick up a new version of the Python driver.


openeo_driver/backend.py
```diff
@@ -295,6 +295,7 @@ class BatchJobMetadata(NamedTuple):
     # TODO: openEO API associates `links` with the job *result* metadata, not the job itself
     links: List[Dict] = None
     usage: Dict = None
+    proj_shape: List[int] = None
```
soxofaan, line 298: FYI: `BatchJobMetadata` has outgrown it's original purpose (pure batch job metadata, not related to batch job results) and I think it needs clean up

About this particular addition: is `proj_shape` batch job metadata or batch job *result* metadata?

openeo_driver/views.py
```diff
@@ -1002,25 +1005,27 @@ def job_result_item_url(item_id) -> str:
...
-                    "stac_extensions": ["eo", "file"],
-                    "id": job_id,
-                    "title": job_info.title,
-                    "description": job_info.description or f"Results for batch job {job_id}",
-                    "license": "proprietary",  # TODO?
-                    "extent": {
-                        "spatial": {
-                            "bbox": [job_info.bbox]
+                result = dict_no_none(
+                    **{
```
soxofaan, line 1009: I think `dict_no_none` also works without those `**`, I'd drop that if you are touching it anyway

openeo_driver/backend.py
JohanKJSchreurs: I agree, I find it confusing too, the difference between BatchJobMetadata and BatchJobResultMetadata should be more clear.

However, I think maybe we should make a separate GitHub issue to fix that properly.
That is a more complicated problem to resolve and it is a different topic than what the current GH issue is about. 

Maybe we can find an easy fix to avoid adding proj_shape in the wrong place for now, and then clean up BatchJobMetadata further in a separate GH issue.

#### Re: proj_shape:

In my understanding proj_shape is really _result_ metadata, because it is info about the assets in the job results.
(extracted via gdalinfo from the asset files) 

So we should find a way to store it with the batch job results, and BatchJobResultMetadata looks like the natural place to put it.

That said, at present the difficulty is specifically in the cases when we want this info at the level of the STAC item:
how to pass that information from the GeopPySpark driver's side into the Python driver.

Also, it would be good to keep epsg, bbox and proj_shape together if possible. 

I am looking into it.

openeo_driver/views.py
JohanKJSchreurs: Done.

openeo_driver/backend.py
JohanKJSchreurs: As discussed this morning, let's merge this for now, so the related PR https://github.com/Open-EO/openeo-geopyspark-driver/pull/405 can move along an be tested. (As well as the additions to the integration tests https://github.com/Open-EO/openeo-geopyspark-integrationtests/pull/5)

After that we can clean up BatchJobMetadata.

soxofaan: > After that we can clean up BatchJobMetadata.

-> https://github.com/Open-EO/openeo-python-driver/issues/190

Issue #187, opened 2023-04-26
Update unit tests that break when Werkzeug updates to v2.3.0


JohanKJSchreurs:
On April 25, 2023 Werkzeug 2.3.0 was released together with Flask 2.3.0.

Werkzeug v2.3.0 caused some unit tests to break:

Namely, In a few tests, the response contains slightly different URLs for the assets `href metadata` field now, because the character `=` was rendered to a safe character sequence `%3D%`  but now the response contains `=` instead.

### Symptoms / logs

```
========================================================================================================= short test summary info ==========================================================================================================
SKIPPED [1] tests/test_views_execute.py:379: parameter checking of callback graphs now happens somewhere else
FAILED tests/test_views.py::TestBatchJobs::test_get_job_results_signed_100 - AssertionError: assert {'assets': {'...n/pdf'}], ...} == {'assets': {'...n/pdf'}], ...}
FAILED tests/test_views.py::TestBatchJobs::test_get_job_results_signed_with_expiration_100 - AssertionError: assert {'assets': {'...n/pdf'}], ...} == {'assets': {'...n/pdf'}], ...}
FAILED tests/test_views.py::TestBatchJobs::test_get_job_results_signed_with_expiration_110 - AssertionError: assert {'assets': {'...fe1f199', ...} == {'assets': {'...fe1f199', ...}
FAILED tests/test_views.py::TestBatchJobs::test_get_job_result_item - AssertionError: assert {'assets': {'...y': None, ...} == {'assets': {'...y': None, ...}
========================================================================================= 4 failed, 1076 passed, 1 skipped, 93 warnings in 14.44s ==========================================================================================
```

For example:

``` 
AssertionError: assert {'assets': {'...n/pdf'}], ...} == {'assets': {'...n/pdf'}], ...}
  Omitting 6 identical items, use -vv to show
  Differing items:
[...]

PR #188, opened 2023-04-26
tests: update unit tests to accept new URLs due to Werkzeug v2.3.0

JohanKJSchreurs:
Fixes https://github.com/Open-EO/openeo-python-driver/issues/187

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -1363,7 +1363,7 @@ class TestBatchJobs:
                     'output.tiff': {
                         'roles': ['data'],
                         'title': 'output.tiff',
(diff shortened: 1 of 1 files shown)


soxofaan:
This PR should also add a minimum constraint `werkzeug>=2.3.0` if it updates the expected outcomes of these tests

soxofaan:
We went for pinning down werkzeug for now in  f27a1ffe799ef6be0656c1988493f227e530edbd

Not sure what we should do with this PR (which now has conflicts).
- just close it 
- update it: with `werkzeug>=2.3.0` constraint, or make the test compatible with both old and new werkzeug

PR #200, opened 2023-06-01
propose run udf on vector cube

jdries:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/437

### DIFF
--- a/openeo_driver/datacube.py
+++ b/openeo_driver/datacube.py
@@ -17,13 +17,14 @@ from pyproj import CRS
 import requests
 
 from openeo.metadata import CollectionMetadata
-from openeo.util import ensure_dir
+from openeo.util import ensure_dir, str_truncate
 from openeo_driver.datastructs import SarBackscatterArgs, ResolutionMergeArgs, StacAsset
 from openeo_driver.errors import FeatureUnsupportedException, InternalException
 from openeo_driver.util.geometry import GeometryBufferer, validate_geojson_coordinates
 from openeo_driver.util.ioformats import IOFORMATS
 from openeo_driver.util.utm import area_in_square_meters
 from openeo_driver.utils import EvalEnv
+from openeogeotrellis.backend import SingleNodeUDFProcessGraphVisitor
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


openeo_driver/datacube.py
```diff
@@ -224,6 +225,38 @@ def with_cube(self, cube: xarray.DataArray, flatten_prefix: str = FLATTEN_PREFIX
...
+    def apply_dimension(
+            self,
+            process: dict,
+            *,
+            dimension: str,
+            target_dimension: Optional[str] = None,
+            context: Optional[dict] = None,
+            env: EvalEnv,
+    ) -> "DriverVectorCube":
+        if dimension == "bands" and target_dimension == None and len(process) == 1 and next(iter(process.values())).get('process_id') == 'run_udf':
```
soxofaan, line 237: Shouldn't this check for the "geometries" dimension instead of the "bands" dimension?

openeo_driver/datacube.py
```diff
@@ -224,6 +225,38 @@ def with_cube(self, cube: xarray.DataArray, flatten_prefix: str = FLATTEN_PREFIX
...
+            visitor = SingleNodeUDFProcessGraphVisitor().accept_process_graph(process)
+            udf = visitor.udf_args.get('udf', None)
+
+            from openeo.udf import FeatureCollection, UdfData
+            collection = FeatureCollection(id='VectorCollection', data=self._as_geopandas_df())
+            data = UdfData(
+                proj={"EPSG": self._geometries.crs.to_epsg()}, feature_collection_list=[collection], user_context=context
```
soxofaan, line 244: only the vector data (labels along geometry dimension) is passed here as feature collection, but the vector cube might also contain actual data cube values

openeo_driver/datacube.py
jdries: I thought so too at first, but when you want to run a udf for each feature, and get all its properties, then the dimension you are working on is "bands" I guess?

openeo_driver/datacube.py
jdries: Looking at _as_geopandas_df, I have the impression that it does merge the geometries with the datacube values?

I'm only a bit worried about from_geojson, which does not seem to extract any property values into cube values, but that's maybe not relevant to what I try to solve here.

openeo_driver/datacube.py
```diff
@@ -17,13 +17,14 @@
...
 from openeo.metadata import CollectionMetadata
-from openeo.util import ensure_dir
+from openeo.util import ensure_dir, str_truncate
 from openeo_driver.datastructs import SarBackscatterArgs, ResolutionMergeArgs, StacAsset
 from openeo_driver.errors import FeatureUnsupportedException, InternalException
 from openeo_driver.util.geometry import GeometryBufferer, validate_geojson_coordinates
 from openeo_driver.util.ioformats import IOFORMATS
 from openeo_driver.util.utm import area_in_square_meters
 from openeo_driver.utils import EvalEnv
+from openeogeotrellis.backend import SingleNodeUDFProcessGraphVisitor
```
soxofaan, line 27: I was trying to build unit test for this feature: but failed on this:
import of geopyspark-driver in python driver code is not going to work

soxofaan:
I iterated further on this feature branch and merged in f83d1cf6f7ea04461bc7a21179f0629dbfee332d

Issue #178, opened 2023-04-17
Include process graph in result metadata (STAC)


soxofaan:
Include process graph in batch job results metadata, see 
- https://github.com/openEOPlatform/architecture-docs/issues/274
- https://github.com/openEOPlatform/architecture-docs/issues/306

PR #202, opened 2023-06-19
Issue178 include process graph in result metadata

JohanKJSchreurs:
Implements https://github.com/Open-EO/openeo-python-driver/issues/178

Need an update of both openeo-python-driver and openeo-geopyspark-driver.

This PR here should be merged before the PR of openeo-geopyspark-driver: 
See: https://github.com/Open-EO/openeo-geopyspark-driver/pull/450


tests/test_views.py
```diff
@@ -40,6 +40,22 @@
 from .data import TEST_DATA_ROOT
 
 
+EXPECTED_PROCESSING_EXPRESSION = [
+    {"expression": {"process_graph": {"foo": {"process_id": "foo", "arguments": {}}}}, "format": "openeo"}
+]
+
+EXPECTED_PROVIDERS = [
+    {
+        "description": "This data was processed on an openEO backend " "maintained by VITO.",
```
jdries, line 49: There seem to be some unneeded quotes here:
backend " "maintained

jdries:
looks good, one minor comment about weird quotes, but otherwise can just be merged

tests/test_views.py
JohanKJSchreurs: Indeed, that was a typo, one that just happens to still work, by accident. 
Hereby corrected.

openeo_driver/views.py
```diff
@@ -919,6 +919,10 @@ def _list_job_results(job_id, user_id):
             job_id=job_id, user_id=user_id
         )
         result_assets = result_metadata.assets
+        providers = result_metadata.providers
+
+        treat_v100_like_v110 = os.environ.get("TREAT_V100_LIKE_V110", "0")
+        TREAT_V100_LIKE_V110 = treat_v100_like_v110.lower() in ["1", "true", "yes"]
```
soxofaan, line 925: FYI we have `smart_bool` for this: https://github.com/Open-EO/openeo-python-driver/blob/bb0e846dfe4c9fa356ce4bf9d062a513f50fe7ad/openeo_driver/utils.py#L152-L157

setup.py
```diff
@@ -24,7 +24,7 @@
     "openeo_udf>=1.0.0rc3",
     "boto3[s3]>=1.26.17",
     "moto>=4.0.10",
-    "time-machine",
+    "time-machine<2.10",  # v2.10 does not provide wheels for Linux (yet) and won't build.
```
soxofaan, line 27: Note I fixed this on master too, which will conflict here

openeo_driver/dummy/dummy_backend.py
```diff
@@ -651,7 +651,30 @@ def get_result_metadata(self, job_id: str, user_id: str) -> BatchJobResultMetada
...
+        job = self.get_job_info(job_id=job_id, user_id=user_id)
+        return [{"expression": job.process, "format": "openeo"}]
+
+    def _get_providers(self, job_id: str, user_id: str):
+        expression = self._get_processing_expression(job_id=job_id, user_id=user_id)
+        return [
+            {
+                "description": "This data was processed on an openEO backend " "maintained by VITO.",
+                "name": "VITO",
+                "processing:expression": expression,
```
soxofaan, line 672: I would avoid VITO references here, and just use "dummy" or something like that

openeo_driver/backend.py
```diff
@@ -426,8 +427,20 @@ def get_result_metadata(self, job_id: str, user_id: str) -> BatchJobResultMetada
...
+    def _get_providers(self, job_id: str, user_id: str):
+        """Creates a standard value for the STAC providers object.
+
+        Helper method for default implementation of `get_result_metadata`.
+
+        For an example implementation, see:
+           openeo_driver.dummy.dummy_backend.DummyBatchJobs._get_providers
+
+        """
+        raise NotImplementedError
```
soxofaan, line 442: Can't we provide a default implementation here, based on https://github.com/Open-EO/openeo-python-driver/blob/bb0e846dfe4c9fa356ce4bf9d062a513f50fe7ad/openeo_driver/config/config.py#L25-L29
which I just added in master coincidentally?

soxofaan: note for tests about this: the dummy backend values are defined here:
https://github.com/Open-EO/openeo-python-driver/blob/bb0e846dfe4c9fa356ce4bf9d062a513f50fe7ad/openeo_driver/dummy/dummy_config.py#L92-L94

soxofaan (commented):
some minor notes

openeo_driver/dummy/dummy_backend.py
JohanKJSchreurs: Updated, along with the solution for one of the comment below: Now it is using the dummy configuration for these test values.

openeo_driver/backend.py
JohanKJSchreurs: Good idea, and it made a few things simpler too.
Implemented it, using the configuration.

PR #208, opened 2023-07-14
Shed importlib resources legacy

### LINKED ISSUE #206: deprecated importlib_resources.path
Integration tests started failing with 
```
  File "/opt/venv/lib64/python3.8/site-packages/openeo_driver/config/load.py", line 76, in get
    self._config = self._load()
  File "/opt/venv/lib64/python3.8/site-packages/openeo_driver/config/load.py", line 87, in _load
    with self._default_config() as default_config:
  File "/opt/venv/lib64/python3.8/site-packages/openeogeotrellis/config/load.py", line 16, in _default_config
    return importlib_resources.path("openeogeotrellis.config", "default.py")
AttributeError: module 'importlib_resources' has no attribute 'path'
```

probably since 6.0.0 release of https://github.com/python/importlib_resources (last week)

jaraco:
- ⚫ Fade to black.
- Switch to non-legacy API for loading resources. Fixes #206.

### DIFF
--- a/openeo_driver/config/load.py
+++ b/openeo_driver/config/load.py
@@ -41,15 +41,11 @@ def load_from_py_file(
     try:
         config = globals[variable]
     except KeyError:
-        raise ConfigException(
-            f"No variable {variable!r} found in config file {path!r}"
-        ) from None
+        raise ConfigException(f"No variable {variable!r} found in config file {path!r}") from None
 
     if expected_class:
         if not isinstance(config, expected_class):
-            raise ConfigException(
-                f"Expected {expected_class.__name__} but got {type(config).__name__}"
-            )
+            raise ConfigException(f"Expected {expected_class.__name__} but got {type(config).__name__}")
     return config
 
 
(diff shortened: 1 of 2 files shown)


jaraco:
I have not tested this change, and I see there aren't CI tests, so please do test the implementation. I would also recommend to import the modules "as" something, e.g. `import importlib_resources as resources` / `from importlib import resources` or to import the names (`as_file`, `files`) directly to make the code more concise, but I retained the current style to minimize the difference.

soxofaan:
Great, thanks!

I cherry-picked the real fix and skipped the "fade to black" as we actually use `darker` instead of `black` to only fixup formatting of changed lines instead of whole files.

I also added a small test to cover actually reading that 'default.py' file.

soxofaan:
about those errors you see in the test run: 
I see "homebrew" so I assume you're running on macos, while we only target linux for now (in testing and deployment). I guess there is some kind dependency side effect going on there.

thanks anyway for contributing!

Issue #215, opened 2023-08-10
support 'partial' query parameter when getting results


jdries:
'partial' query parameter:
https://api.openeo.org/#tag/Data-Processing/operation/list-results

So when partial is True, the result should mainly contain:
"openeo:status": "running"
Setting other metadata elements would of course be nice, but not sure to what extent this is possible.

We also need to add the partial parameter back to the canonical (signed) url.

bossie:
I realize that this is still in progress but I've been thinking about some edge cases in the context of https://github.com/Open-EO/openeo-geopyspark-driver/issues/489  that might not have been covered yet at this point:

1. if partial and job error/canceled -> openeo:status running (expected: error/canceled)
2. if partial and job finished -> no openeo:status (expected: finished)
3. if not partial and job error/canceled -> JobNotFinishedException; after some contemplating I believe this is still OK as the back-end is not required to return intermediate results.

PR #216, opened 2023-08-10
Issue215 support partial query parameter when getting results

JohanKJSchreurs:
Implements: https://github.com/Open-EO/openeo-python-driver/issues/215


jdries:
One thing that is important to include is the 'canonical' link in the links section, which comes down to this piece of code
```

 {
                      "rel": "canonical",
                      "href": job_results_canonical_url(),
                      "type": "application/json",
                  }
```

PR #224, opened 2023-09-08
Issue #215 Fix: report correct job status in _list_job_results when p…

JohanKJSchreurs:
## Issue #215 Fix: report correct job status in _list_job_results when partial=true

This PR corrects the reported job status for some cases in the context of https://github.com/Open-EO/openeo-geopyspark-driver/issues/489, as noted by  @bossie, here: https://github.com/Open-EO/openeo-python-driver/issues/215#issuecomment-1682261335 .

- if partial and job error/canceled -> openeo:status running (expected: error/canceled)
- if partial and job finished -> no openeo:status (expected: finished)
- if not partial and job error/canceled -> JobNotFinishedException; after some contemplating I believe this is still OK as the back-end is not required to return intermediate results.

### Solution in brief:

Basically we should report the actual job status when partial is true, and also include openeo:status when the job is finished. 
- When partial is false and the jobs is not finished, do the same as before: raise JobNotFinishedException.

### DIFF
--- a/openeo_driver/views.py
+++ b/openeo_driver/views.py
@@ -956,9 +956,11 @@ def register_views_batch_jobs(
                 elif job_info.status == JOB_STATUS.CANCELED:
                     openeo_status = "canceled"
                 elif job_info.status == JOB_STATUS.QUEUED:
-                    openeo_status = "queued"
+                    openeo_status = "running"
                 elif job_info.status == JOB_STATUS.CREATED:
-                    openeo_status = "created"
+                    openeo_status = "running"
+                else:
+                    raise AssertionError(f"unexpected job status: {job_info.status!r}")
 
                 result = {
                     "openeo:status": openeo_status,
--- a/openeo_driver/views.py
+++ b/openeo_driver/views.py
@@ -949,8 +949,19 @@ def register_views_batch_jobs(
             if not partial:
                 raise JobNotFinishedException()
             else:
+                if job_info.status == JOB_STATUS.RUNNING:
+                    openeo_status = "running"
+                elif job_info.status == JOB_STATUS.ERROR:
+                    openeo_status = "error"
+                elif job_info.status == JOB_STATUS.CANCELED:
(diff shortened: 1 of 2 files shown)


openeo_driver/views.py
```diff
@@ -949,8 +949,19 @@ def job_results_canonical_url() -> str:
...
+                if job_info.status == JOB_STATUS.RUNNING:
+                    openeo_status = "running"
+                elif job_info.status == JOB_STATUS.ERROR:
+                    openeo_status = "error"
+                elif job_info.status == JOB_STATUS.CANCELED:
+                    openeo_status = "canceled"
+                elif job_info.status == JOB_STATUS.QUEUED:
+                    openeo_status = "queued"
+                elif job_info.status == JOB_STATUS.CREATED:
+                    openeo_status = "created"
```
bossie, line 961: According to https://openeo.org/documentation/1.0/developers/api/reference.html#tag/Batch-Jobs/operation/list-results, the allowed values for "openeo:status" are: "running" "canceled" "finished" "error" so you'll want to map job statuses "queued" and "created" to "openeo:status" "running" instead.

openeo_driver/views.py
```diff
@@ -949,8 +949,19 @@ def job_results_canonical_url() -> str:
...
+
                 result = {
-                    "openeo:status": "running",
+                    "openeo:status": openeo_status,
```
bossie, line 964: I'm getting this warning in Pycharm:

> Local variable 'openeo_status' might be referenced before assignment

so maybe add something like this:
```
else:
    raise AssertionError(f"unexpected job status {job_info.status}")
```

bossie (changes requested):
Added remarks inline.

openeo_driver/views.py
JohanKJSchreurs: Added the exception, and added a test for it as well.

bossie (approved):
Looks great.

Issue #214, opened 2023-08-09
include web app driver errors in batch job logs


bossie:
A batch job with this (faulty) process graph will trigger an error in the web app driver upon starting the job (the actual error is not important):

```json
{
  "process_graph": {
    "load1": {
      "arguments": {
        "bands": [
          "B03",
          "B04"
        ],
        "id": "TERRASCOPE_S2_TOC_V2",
        "spatial_extent": [1,2,3,4],
        "temporal_extent": ["2021-07-01","2021-07-20"]
      },
      "process_id": "load_collection",
      "result": true
    }
  }
}
```

```
Traceback (most recent call last):
  File "/opt/venv/lib64/python3.8/site-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/opt/venv/lib64/python3.8/site-packages/flask/app.py", line 1502, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/opt/venv/lib64/python3.8/site-packages/openeo_driver/users/auth.py", line 88, in decorated
    return f(*args, **kwargs)
  File "/opt/venv/lib64/python3.8/site-packages/openeo_driver/views.py", line 866, in queue_job
    backend_implementation.batch_jobs.start_job(job_id=job_id, user=user)
  File "/opt/venv/lib64/python3.8/site-packages/openeogeotrellis/backend.py", line 1435, in start_job
    self._start_job(job_id, user, _get_vault_token)
  File "/opt/venv/lib64/python3.8/site-packages/openeogeotrellis/backend.py", line 1473, in _start_job
    and self._scheduled_sentinelhub_batch_processes(
  File "/opt/venv/lib64/python3.8/site-packages/openeogeotrellis/backend.py", line 1900, in _scheduled_sentinelhub_batch_processes
    convert_node(result_node, env=env.push({ENV_DRY_RUN_TRACER: dry_run_tracer, ENV_SAVE_RESULT:[],"node_caching":False}))
  File "/opt/venv/lib64/python3.8/site-packages/openeo_driver/ProcessGraphDeserializer.py", line 368, in convert_node
    process_result = apply_process(process_id=process_id, args=processGraph.get('arguments', {}),
  File "/opt/venv/lib64/python3.8/site-packages/openeo_driver/ProcessGraphDeserializer.py", line 1512, in apply_process
[...]

PR #233, opened 2023-10-18
Issue #214 automatically log extra data with a context manager

soxofaan:
Issue #214 proof of concept to automatically log extra data with a context manager


openeo_driver/util/logging.py
```diff
@@ -345,6 +345,51 @@ def filter(self, record: logging.LogRecord) -> bool:
         return True
 
 
+class ContextBasedExtraInjectingFilter(logging.Filter):
```
soxofaan, line 348: maybe just `ExtraLoggingFilter` to keep it short

bossie:
@soxofaan does my test cover the threading behaviour?

soxofaan:
nice, one thing  that might be added to that test is some way to make sure that there is (forced) context switching between the threads, because now a passing test could just be the result of each thread finishing before the next has started

openeo_driver/util/logging.py
soxofaan: done (also leaves open the option to cover non-context-manager use cases)

soxofaan:
rebased and merged in e276c325de6f9e4e8c7e362a70534d61a5407336

PR #244, opened 2024-01-05
phase out ZKJobRegistry

bossie:
Implements part of https://github.com/Open-EO/openeo-geopyspark-driver/issues/632.


bossie (commented):
Not a fan of the  naming of `openeo_driver.jobregistry.JobRegistryInterface.set_results_metadata`; it's ultimately called by JobTracker upon completion of a job.

bossie:
TODO:

- [ ] revert package name temporarily changed to be able to reference it in the corresponding openeo_geopyspark build

Jenkinsfile
```diff
@@ -15,7 +15,7 @@ pythonPipeline {
   downstream_job = 'openEO/openeo-integrationtests'
   wheel_repo = 'python-openeo'
   extras_require = 'dev'
-  upload_dev_wheels = false
+  upload_dev_wheels = true
```
bossie, line 18: Revert this.

setup.py
```diff
@@ -31,7 +31,7 @@
 ]
 
 setup(
-    name="openeo_driver",
+    name="openeo_driver_phase_out_zkjobregistry",
```
bossie, line 34: Revert this.

tests/test_server.py
```diff
@@ -8,5 +8,5 @@ def test_build_backend_deploy_metadata():
     data = build_backend_deploy_metadata(packages=["openeo", "openeo_driver", "foobarblerghbwop"])
     assert data["date"].startswith(datetime.utcnow().strftime("%Y-%m-%dT%H"))
     assert re.match(r"\d+\.\d+\.\d+", data["versions"]["openeo"])
-    assert re.match(r"\d+\.\d+\.\d+", data["versions"]["openeo_driver"])
+    # assert re.match(r"\d+\.\d+\.\d+", data["versions"]["openeo_driver"])
```
bossie, line 11: Revert this.

tests/test_views.py
```diff
@@ -276,10 +276,10 @@ def test_capabilities_oidc_auth_default(self, enable_oidc_auth):
...
     def test_capabilities_processing_software(self, api100):
         capabilities = api100.get('/').assert_status_code(200).json
-        assert capabilities["processing:software"] == {
-            "openeo": RegexMatcher(r"0.\d+\.\d+"),
-            "openeo_driver": RegexMatcher(r"0.\d+\.\d+"),
-        }
+        # assert capabilities["processing:software"] == {
+        #     "openeo": RegexMatcher(r"0.\d+\.\d+"),
+        #     "openeo_driver": RegexMatcher(r"0.\d+\.\d+"),
+        # }
```
bossie, line 282: Revert this.

bossie (commented):
Revert stuff temporarily introduced to get the entire thing to build.

openeo_driver/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.80.0a1"
+__version__ = "0.81.0a1"
```
soxofaan, line 1: FYI: I bumped this already in 26a0612e72cb9a7105acbf4943d090ee64cd5f76  so you want to bump it more :)

openeo_driver/jobregistry.py
```diff
@@ -118,9 +119,15 @@ def set_proxy_user(self, job_id: str, proxy_user: str) -> JobDict:
     def set_application_id(self, job_id: str, application_id: str) -> JobDict:
         raise NotImplementedError
 
+    @deprecated("call set_results_metadata instead", version="0.81.0")
```
soxofaan, line 122: To be verified, but I think adding `deprecated` on "unimplemented" base class methods has no practical effect. Still it has documentation value of course

openeo_driver/jobregistry.py
```diff
@@ -118,9 +119,15 @@ def set_proxy_user(self, job_id: str, proxy_user: str) -> JobDict:
...
     def set_usage(self, job_id: str, costs: float, usage: dict) -> JobDict:
         raise NotImplementedError
 
+    # TODO: improve name?
+    def set_results_metadata(self, job_id: str, costs: Optional[float], usage: dict,
```
soxofaan, line 127: It's probably a style things, but if you declare `costs: Optional[float]` you should give it a default `None` value otherwise it's not actually optional. 
If something is not optional, but still `None`-able, I often use `Union[None, float]` instead.
Both are technically equivalent, but their documenting nature is different

openeo_driver/jobregistry.py
```diff
@@ -139,12 +146,18 @@ def list_active_jobs(
...
 
         :param max_age: optional filter to only return recently created jobs:
-            maximum number of days creation date.
+            creation date is at most max_age days ago.
         :param fields: job metadata fields that should be included in result
         """
         # TODO: option for job metadata fields that should be included in result
         raise NotImplementedError
 
+    def list_trackable_jobs(self, fields: Optional[List[str]] = None) -> List[JobDict]:
```
soxofaan, line 155: Reading this PR top to bottom, so answer is probably below, but what is `fields` for? Fields that should be set in the returned listing? Or extra fields to add on top op some defaults?

openeo_driver/jobregistry.py
```diff
@@ -216,6 +229,9 @@ def __init__(
         session: Optional[requests.Session] = None,
         _debug_show_curl: bool = False,
     ):
+        if api_url is None:
```
soxofaan, line 232: you could even make this 
```suggestion
        if not api_url:
```
to also fail on empty strings (or other "falsy" values)

soxofaan (approved):
apart from some small notes, this PR mainly adds new APIs, leaving original logic basically alone, so looks fine to me

openeo_driver/jobregistry.py
bossie: Didn't know it actually did something. But it's for documentation purposes yeah.

openeo_driver/jobregistry.py
bossie: I considered `Optional` == the value can be `None`,  not: it has a default value so you don't have to pass it explicitly to the method call.

This is explicitly mentioned in the [docs](https://docs.python.org/3/library/typing.html#typing.Optional), no?

> if an explicit value of None is allowed, the use of Optional is appropriate, whether the argument is optional or not.

openeo_driver/jobregistry.py
bossie: It mirrors the `fields` argument of its sibling methods. I suppose it means: you're sure to get these fields back (and some more).

openeo_driver/jobregistry.py
soxofaan: FYI: calling  a `deprecate`d function will trigger a UserWarning or something like that

openeo_driver/jobregistry.py
soxofaan: yes never mind, this is purely a cosmetic thing I heard some debate about recently in the python community. Not something to stall this PR on

PR #258, opened 2024-02-07
Implement verify_for_synchronous_processing. 

EmileSonneveld:
Only show a warning for the moment. When we have it running for a while, we can evaluate if it does not give too many false positives and then we can throw warning instead.
https://github.com/Open-EO/openeo-geopyspark-driver/issues/616


EmileSonneveld:
@soxofaan Is `ConcreteProcessing` a good place to implement `verify_for_synchronous_processing`?

openeo_driver/views.py
```diff
@@ -647,10 +647,10 @@ def result(user: User):
                 _log.exception(f"Unexpected error while verifying synchronous processing: {e}")
             else:
                 if sync_processing_issues:
-                    raise ProcessGraphComplexityException(
-                        message=ProcessGraphComplexityException.message
-                        + f" Reasons: {' '.join(sync_processing_issues)}"
-                    )
+                    # After this proves to run correctly on many sync jobs, we can throw an error instead.
+                    _log.warning(ProcessGraphComplexityException.message
+                                 + f" Reasons: {' '.join(map(lambda x: str(x), sync_processing_issues))}"
```
soxofaan, line 652: this `map(lambda x: str(x)` should not be necessary as sync_processing_issues is a list of strings

openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -340,6 +340,17 @@ def extra_validation(
...
+    def verify_for_synchronous_processing(self, process_graph: dict, env: EvalEnv = None):
+        env_validate = env.push({
+            "allow_check_missing_products": False,
+            "sync_job": True,
+        })
+        errors = self.validate(process_graph=process_graph, env=env_validate)
+
+        # Only care for certain errors:
+        errors = list(filter(lambda x: x["code"] == "ExtentTooLarge", errors))
+        return errors
```
soxofaan, line 352: this violates the return type of verify_for_synchronous_processing, which should return a list of strings, while here it returns a list of dicts

Related, I think it's easier to use list comprehension instead of `list(filter(lambda` constructs:

```
errors = [e["message"] for e in errors if e["code"] == "ExtentTooLarge"]
```

soxofaan:
>  @soxofaan Is ConcreteProcessing a good place to implement verify_for_synchronous_processing?

It feels a bit wrong, because:
- ConcreteProcessing is also used in the aggregator, which has different constraints than the geopyspark deploy
- Current PR filters on "ExtentTooLarge", which is only defined in openeo-geopyspark-driver, but nowhere used in openeo-python-driver

I think it should be implemented in `GpsProcessing`

openeo_driver/ProcessGraphDeserializer.py
EmileSonneveld: fixed

EmileSonneveld:
`GpsProcessing` makes sense indeed. It is also where `extra_validation` is implemented.
Moved

openeo_driver/views.py
EmileSonneveld: true, fixed

PR #267, opened 2024-03-04
663 aggregate spatial can not be used as input to vector to raster 1

JeroenVerstraelen:


openeo_driver/save_result.py
```diff
@@ -533,6 +544,34 @@ def to_geoparquet(self, destination: Optional[str] = None) -> str:
         gpd.GeoDataFrame(stats.join(gdf, on='feature_index')).to_parquet(filename)
         return filename
 
+    def to_driver_vector_cube(self) -> DriverVectorCube:
+        if isinstance(self._regions, GeometryCollection):
```
soxofaan, line 548: FYI, I guess this is a legacy support thing, but a GeoJSON `GeometryCollection` is strictly speaking a single, indivisible "feature" that should not be split up in its sub-geometries

openeo_driver/datacube.py
```diff
@@ -473,6 +474,17 @@ def _as_geopandas_df(
             # TODO: better way to combine cube with geometries
             # Flatten multiple (non-geometry) dimensions from cube to new properties in geopandas dataframe
             if self._cube.dims[1:]:
+                if self.DIM_TIME in self._cube.dims:
+                    # Ensure dates are in 2017-10-25T11:37:00Z format.
+                    time_coords: list[Any] = list(self._cube.coords[self.DIM_TIME].values)
+                    try:
+                        # TODO: Support datetime, pandas.Timestamp, numpy.datetime64, etc?
+                        self._cube.coords[self.DIM_TIME] = [
+                            pandas.to_datetime(coord).tz_convert("UTC").strftime("%Y-%m-%dT%H:%M:%SZ")
```
soxofaan, line 483: this looks like something that deserves a couple of unit tests

openeo_driver/save_result.py
```diff
@@ -732,6 +771,26 @@ def fit_class_random_forest(
         # TODO: this method belongs eventually under DriverVectorCube
         raise NotImplementedError
 
+    def get_data(self) -> gpd.GeoDataFrame:
+        if isinstance(self._regions, DriverVectorCube):
+            gdf = self._regions._geometries
```
soxofaan, line 776: I would avoid using private attr `_geometries` here

openeo_driver/save_result.py
```diff
@@ -732,6 +771,26 @@ def fit_class_random_forest(
         # TODO: this method belongs eventually under DriverVectorCube
         raise NotImplementedError
 
+    def get_data(self) -> gpd.GeoDataFrame:
```
soxofaan, line 774: I think it's risky to define this as `get_data`, which is also defined in comparable classes, but with different return type. As far as I understand, this method is only going to be used in to_driver_vector_cube, so it could be an internal `_get_data()` instead

openeo_driver/save_result.py
```diff
@@ -732,6 +771,26 @@ def fit_class_random_forest(
...
+        elif isinstance(self._regions, GeometryCollection):
+            gdf = gpd.GeoDataFrame(geometry=list(self._regions.geoms))
+        elif isinstance(self._regions, BaseGeometry):
+            gdf = gpd.GeoDataFrame(geometry=[self._regions])
+        else:
+            raise NotImplementedError
+        gdf["feature_index"] = gdf.index
+        stats: pd.DataFrame = pd.read_csv(self._csv_path())
+        gdf = gdf.join(stats.set_index("feature_index"), on="feature_index").rename(
+            columns=lambda col_name: re.sub(r"\W", "_", col_name)
```
soxofaan, line 786: what is the reason for the `\W` rename?

soxofaan (commented):
some quick notes

soxofaan:
overall I think there should be more unit test coverage because there are a lot of small details to this

PR #278, opened 2024-03-25
Throw exception when verify_for_synchronous_processing gives errors. 

EmileSonneveld:


openeo_driver/views.py
```diff
@@ -648,7 +648,7 @@ def result(user: User):
             else:
                 if sync_processing_issues:
                     # After this proves to run correctly on many sync jobs, we can throw an error instead.
-                    _log.warning(ProcessGraphComplexityException.message
+                    raise OpenEOApiException(ProcessGraphComplexityException.message
```
soxofaan, line 651: You should raise a `ProcessGraphComplexityException` exception (instead of the generic `OpenEOApiException`), which automatically has the right properties (like status code and error code

PR #279, opened 2024-03-27
276 aggregate spatial and vector to raster return aggregatepolygonresult in dry run

JeroenVerstraelen:


openeo_driver/dry_run.py
```diff
@@ -556,10 +556,7 @@ def aggregate_spatial(
             geometries, bbox = self._normalize_geometry(geometries)
             cube = self.filter_bbox(**bbox, operation="_weak_spatial_extent")
         cube._process(operation="aggregate_spatial", arguments={"geometries": geometries})
-        if isinstance(geometries, (Polygon, MultiPolygon)):
-            # TODO #71 normalize to feature collection instead of deprecated geometry collection
-            geometries = GeometryCollection([geometries])
-        return AggregatePolygonResult(timeseries={}, regions=geometries)
+        return cube
```
soxofaan, line 559: shouldn't the line above be as follows?

```
cube = cube._process(....
```

or together with the return  like in other places here:

```
return cube._proces(...
```

PR #290, opened 2024-06-10
Add message to assert, and add it to the ProcessGraphInvalidException

EmileSonneveld:


openeo_driver/errors.py
```diff
@@ -371,6 +371,12 @@ class ProcessGraphInvalidException(OpenEOApiException):
     _description = "The process doesn't contain a valid process graph, which means it doesn't comply to the general structure / schema."
     _tags = ['Batch Jobs', 'Data Processing', 'Secondary Services', 'User-Defined Processes']
 
+    def __init__(self, message: str = ""):
```
soxofaan, line 374: I would not add methods here, the exception classes here are compiled/autogenerated from JSON from openeo-api:

https://github.com/Open-EO/openeo-python-driver/blob/6749cdee1bff04fed8c71ba347d89588b6d148c4/openeo_driver/errors.py#L87


Note that the OpenEOApiException base class already has an `__init__` that should be enough to do this.
And also note that by overriding `__init__` you'll probably break current usage of that

openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -1871,12 +1871,13 @@ def evaluate_process_from_url(process_id: str, namespace: str, args: dict, env:
 
     try:
         spec = res.json()
-        assert spec["id"].lower() == process_id.lower()
+        msg = f"{process_id=} does not match result from namespace query: '" + spec["id"] + "'"
```
soxofaan, line 1874:
```suggestion
        msg = f"Mismatch between expected process {process_id!r} and process {spec['id']!r} defined at {candidate!r}.'"
```

openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -1871,12 +1871,13 @@ def evaluate_process_from_url(process_id: str, namespace: str, args: dict, env:
...
+        assert spec["id"].lower() == process_id.lower(), msg
         process_graph = spec["process_graph"]
         parameters = spec.get("parameters", [])
     except Exception as e:
         _log.error(f"Failed to load process {process_id!r} from {candidate!r}", exc_info=True)
-        raise ProcessGraphInvalidException() from e
+        raise ProcessGraphInvalidException(str(e)) from e
```
soxofaan, line 1880:
```suggestion
        raise ProcessGraphInvalidException(message=str(e)) from e
```

openeo_driver/errors.py
EmileSonneveld: The 'Invalid process graph specified.' message is gone then. Ok?

PR #293, opened 2024-06-13
Early check if UDF returns valid polygons to give more clear error. 

EmileSonneveld:
@soxofaan Implement this check here, or better in gepyspark-driver? https://github.com/Open-EO/openeo-geopyspark-driver/blob/7b4fbe9ceb78f1e778ebef15a1b00c786e07bc2f/openeogeotrellis/udf.py#L33

### DIFF
--- a/openeo_driver/ProcessGraphDeserializer.py
+++ b/openeo_driver/ProcessGraphDeserializer.py
@@ -1456,8 +1456,12 @@ def run_udf(args: dict, env: EvalEnv):
             dataframe = gpd.GeoDataFrame(geometry=geo_data)
         invalid_indexes = dataframe.index[~dataframe.is_valid].tolist()
         if len(invalid_indexes) > 0:
-            raise ValueError("UDF returned invalid polygons. This could "
-                             + f"be due to the input or the code. Invalid index(es): {invalid_indexes}")
+            raise OpenEOApiException(
+                status_code=400,
+                code="InvalidGeometry",
+                message="UDF returned invalid polygons. This could "
+                        + f"be due to the input or the code. Invalid index(es): {invalid_indexes}"
+            )
         return DriverVectorCube.from_geodataframe(data=dataframe)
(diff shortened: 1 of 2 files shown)


openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -1454,6 +1454,10 @@ def run_udf(args: dict, env: EvalEnv):
         dataframe = geo_data
         if isinstance(geo_data, gpd.GeoSeries):
             dataframe = gpd.GeoDataFrame(geometry=geo_data)
+        invalid_indexes = dataframe.index[~dataframe.is_valid].tolist()
+        if len(invalid_indexes) > 0:
+            raise ValueError("UDF returned invalid polygons. This could "
```
soxofaan, line 1459: If you raise an exception with a message for the user to see, you should raise a `OpenEOApiException` (preferably one of its subclasses, if applicable), otherwise the user will receive a generic "500 Internal server error"

In this case it's a user error, so it should at least be a status 400. There is not really a predefined error code, but you can invent one and specify that:

```suggestion
            raise OpenEOApiException(
                status_code=400,
                code="InvalidGeometry",
                message="UDF produced invalid geometry. This could "
```

soxofaan (commented):
good idea, 

I'm even wondering if this shouldn't be actually moved to the `DriverVectorCube` constructor

openeo_driver/ProcessGraphDeserializer.py
EmileSonneveld: I applied this and added a warning to the driver vector cube constructor. I'll merge to master already to see how the tests go

EmileSonneveld: No `DriverVectorCube with invalid polygon` messages in python-driver nor geopyspark-driver nor integration test logs found. But still I don't dare to throw an actual exception. Maybe after checking the kibana logs in a few weeks

PR #301, opened 2024-07-23
Stac alternate

EmileSonneveld:
Proxy `requests.models.Request` to use the existing `urllib` mocking

### DIFF
--- a/openeo_driver/testing.py
+++ b/openeo_driver/testing.py
@@ -19,6 +19,7 @@ import attrs
 import openeo
 import openeo.processes
 import pytest
+import requests
 import shapely.geometry.base
 import shapely.wkt
 from flask import Response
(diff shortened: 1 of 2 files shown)


bossie:
Is this about not having to think about which HTTP library is actually being used underneath e.g. `load_stac` that uses both `requests` and `urllib`?

EmileSonneveld:
Exactly. The load stac uses both ways internally. So had to extend the testing proxy to make simple tests

bossie (commented):
I have some issues with this approach:

- it duplicates tried and tested `requests_mock` functionality;
- there's a naming problem: `UrllibMocker` is specific to `urllib`.

A solution might be to move this to a utility function (or similar, I'm not sure what this looks like exactly, implementation-wise) instead of adapting `UrllibMocker` i.e. treat this as if we don't have access to the `UrllibMocker` source.

Also: this only supports GET requests but I guess that can be easily fixed.

PR #310, opened 2024-09-05
7 calculation of zonal statistics in this case zonal maximum in openeo

bossie:
https://github.com/EU-GRASSLAND-WATCH/EUGW/issues/7


bossie:
Allows running the process graph to completion by:

- supporting passing the output of `raster_to_vector` to `aggregate_spatial` during the dry run **(quick fix)**,
- making sure that `vector_to_raster` passes GeoJSON in EPSG:4326 as expected by [the implementation in geotrellis-extensions](https://github.com/Open-EO/openeo-geotrellis-extensions/blob/5650f36533ccb29440bab35beaa9c95aeb3d2eef/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/vector/VectorCubeMethods.scala#L72).

PR #312, opened 2024-09-16
default start/end_datetime for STAC Items

bossie:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/852


bossie:
Fixed by populating actual `start_datetime`/`end_datetime` (https://github.com/Open-EO/openeo-geopyspark-driver/pull/864) instead of falling back to a default.

PR #314, opened 2024-09-25
add support for apply_vectorcube udf Open-EO/openeo-geopyspark-driver…

JeroenVerstraelen:
…#881


openeo_driver/datacube.py
```diff
@@ -770,9 +771,45 @@ def apply_dimension(
...
+                                f"UDF should return single datacube but got {datacube_list and len(datacube_list)}"
+                            )
+                        return feature_list[0].data, datacube_list[0].array
+                    return feature_list[0].data, None
+
                 log.warning(
                     f"Using experimental feature: DriverVectorCube.apply_dimension along dim {dimension} and empty cube"
                 )
+                regex = re.compile(r"^def apply_vectorcube\(")
+                if any(regex.match(line) for line in single_run_udf.udf.splitlines()):
```
soxofaan, line 796: you don't have to split the UDF in lines, you can just do the following I think
```suggestion
                if regex.match(single_run_udf, flags=re.MULTILINE):
```

soxofaan:
also consider adding tests,

e.g. see
https://github.com/Open-EO/openeo-python-driver/blob/b2c20b5cc55baa41cb354c10be535a5fc3c63788/tests/test_vectorcube.py#L882-L988
and
https://github.com/Open-EO/openeo-python-driver/blob/b2c20b5cc55baa41cb354c10be535a5fc3c63788/tests/test_views_execute.py#L4101-L4377

openeo_driver/datacube.py
JeroenVerstraelen: It doesn't look like flags is a supported argument for regex.match.

soxofaan: ah yeah, that's an annoying thing with `re` in python, it does not support keyword args, you have to do it positional: 
```suggestion
                if regex.match(single_run_udf, re.MULTILINE):
```

JeroenVerstraelen: I tried all variations for regex.match but it seems like it's not supported:
```
regex = re.compile(r"^def apply_vectorcube\(")
found = regex.match(single_run_udf, re.MULTILINE)
```
```
regex = re.compile(r"^def apply_vectorcube\(", re.MULTILINE)
found = regex.match(single_run_udf)
```

It does work with re.findall however:
`re.findall(r"^def apply_vectorcube\(", single_run_udf, re.MULTILINE)`

soxofaan: ok sorry I should not have shared code I didn't try

this works for me:
```python
import re
regex = re.compile("^def foo", re.MULTILINE)

print(regex.search("""
#bla
# def foo
"""))
print(regex.search("""
bla
def foo
"""))
```

openeo_driver/datacube.py
```diff
@@ -770,9 +771,44 @@ def apply_dimension(
...
+                            raise ValueError(
+                                f"UDF should return single datacube but got {datacube_list and len(datacube_list)}"
+                            )
+                        return feature_list[0].data, datacube_list[0].array
+                    return feature_list[0].data, None
+
                 log.warning(
                     f"Using experimental feature: DriverVectorCube.apply_dimension along dim {dimension} and empty cube"
                 )
+                if len(re.findall(r"^def apply_vectorcube\(", single_run_udf.udf, re.MULTILINE)) != 0:
```
soxofaan, line 795: so this should be more to the point (with some additional whitespace paranoia):
```suggestion
                if re.search(f"^def\s+apply_vectorcube\s*\(", single_run_udf.udf, re.MULTILINE):
```

openeo_driver/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.107.7a1"
+__version__ = "0.108.0a1"
```
soxofaan, line 1: I would also add a changelog entry for tis

soxofaan:
and the client dependency in setup.py
https://github.com/Open-EO/openeo-python-driver/blob/af0133802f8573405d9d17b7dbaec7450ef4c2da/setup.py#L56

should be bumped to properly signal the dependency on https://github.com/Open-EO/openeo-python-client/pull/631
I think it should become `"openeo>=0.32.0.a2.dev"`

soxofaan:
follow up:
- https://github.com/Open-EO/openeo-python-driver/issues/315

Issue #323, opened 2024-11-07
dry_run of run_udf process only produces JSONResult


JeroenVerstraelen:
A dry run of the `run_udf` process can only generate a `JSONResult` object, but a wet run is also able to produce a DriverVectorCube.

PR #324, opened 2024-11-07
ensure run_udf returns a DryRunDataCube if required #323

JeroenVerstraelen:


soxofaan (commented):
fix looks even simpler than we discussed earlier, so that's good.

just some unit test notes

soxofaan (approved):
good to merge (if tests pass)

but I would also add a changelog entry (and maybe version bump)

PR #328, opened 2024-11-22
auto-convert incoming geojson to vectorcube

jdries:
For consistent and correct handling of geometry in our backend, we need vector cubes.
This code path was still emitting shapely geometry objects, for no clear/good reason, causing unexpected behaviour:
https://github.com/VITO-RS-Vegetation/lcfm-production/issues/88

The error logs show clearly that geometry with different CRS's was incorrectly combined when vector cubes are not used.

Next to that, annoying messages were printed in the logs:

```
geojson_to_geometry usage is deprecated and should be replaced by proper vector cube usage
_normalize_geometry: TODO are we still reaching this code?
```

This pr suggests to get rid of all that via a simple translation to vector cube, which is a code path with much better support and normally also test coverage.

@soxofaan  @JeroenVerstraelen  I can add a unit test, but would first like to learn if there were reasons to not do this?

### DIFF
--- a/openeo_driver/dry_run.py
+++ b/openeo_driver/dry_run.py
@@ -593,11 +593,6 @@ class DryRunDataCube(DriverDataCube):
         elif isinstance(geometries, DelayedVector):
             bbox = geometries.bounds
         elif isinstance(geometries, shapely.geometry.base.BaseGeometry):
-            _log.warning(
-                "_normalize_geometry: TODO are we still reaching this code?",
-                # TODO yes we are, apparently due to #288
-                stack_info=True,
-            )
             # TODO: buffer distance of 10m assumes certain resolution (e.g. sentinel2 pixels)
             # TODO: use proper distance for collection resolution instead of using a default distance?
             # TODO: or eliminate need for buffering in the first place? https://github.com/Open-EO/openeo-python-driver/issues/148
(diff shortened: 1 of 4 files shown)


jdries:
Tests fixed, coverage seemed to be good already. I'll merge if no objections?

openeo_driver/dry_run.py
```diff
@@ -593,11 +593,6 @@ def _normalize_geometry(self, geometries) -> Tuple[Union[DriverVectorCube, Delay
         elif isinstance(geometries, DelayedVector):
             bbox = geometries.bounds
         elif isinstance(geometries, shapely.geometry.base.BaseGeometry):
-            _log.warning(
-                "_normalize_geometry: TODO are we still reaching this code?",
-                # TODO yes we are, apparently due to #288
-                stack_info=True,
-            )
```
soxofaan, line 600: can't we keep this warning to make sure this code path has effectively died in practice?

Or should we change the message to make it more actionable for users?

jdries: can't keep it in this form, it prints a warning for something the user doesn't know about, and the message includes a 'TODO'.
Also a simple usage search also told me that this code path was very much still reachable. I'm not against using logging to detect the same thing, but only if we then also take action.

PR #339, opened 2024-12-11
Avoid exception for common logic. 

EmileSonneveld:
Got this while debugging with break on raise exception

### DIFF
--- a/openeo_driver/utils.py
+++ b/openeo_driver/utils.py
@@ -51,9 +51,9 @@ class EvalEnv:
             raise KeyError(key)
 
     def get(self, key: str, default=None) -> Any:
-        try:
+        if key in self:
             return self[key]
-        except KeyError:
+        else:
             return default
 
     def push(self, values: dict = None, **kwargs) -> 'EvalEnv':


EmileSonneveld:
Configured like this, PyCharm wil breakpoint the code wight on the moment an exception occurs:
image

```python
not issubclass(__exception__[0], ImportError) and not issubclass(__exception__[0], GeneratorExit)
```

soxofaan:
I'm fine with merging this, but I wonder if that really improves your debug experience, as "ask forgiveness instead of permission" (like this try-except-KeyError) is a pretty common idiom used in a lot of other places/libs.

EmileSonneveld:
Adding conditionals to debugging code has a big performance hit. Removing any common exception speeds it up

EmileSonneveld:
I used to love it :) Python is a bit more exception happy than other languages, but when ignoring external libraries, even in Python caught exceptions mean something might be going on. Like the Internal exceptions in the extra_validation. Or it can early detect HTTP calls being retried without logging.

EmileSonneveld:
Another error where the break on exception is handy for me. This exception gets ignored when parsing stac metadata, only to crash later-on:
`ValueError: Invalid band name/index 'B02'. Valid names: []`

Issue #342, opened 2024-12-16
timeseries.json should be written in job folder


EmileSonneveld:
CDSE-196
Got this error: `No such key: s3://OpenEO-data/batch_jobs/timeseries.json`

PR #343, opened 2024-12-16
Suffix to remove

EmileSonneveld:
https://github.com/Open-EO/openeo-python-driver/issues/342

There are 2 `write_assets` functions. Both takes the suffix from the paths. One can call the other, but this happens after the suffix is removed from the path. Here, I add a new suffix so that the nested `write_assets` works fine

### DIFF
--- a/openeo_driver/datacube.py
+++ b/openeo_driver/datacube.py
@@ -576,18 +576,18 @@ class DriverVectorCube:
             options = {}
         # TODO: We should treat the directory parameter as an actual directory. (Open-EO/openeo-geopyspark-driver#888)
         filename = str(directory)
-        directory = Path(filename).parent
+        directory_parent = Path(filename).parent
 
         format_info = IOFORMATS.get(format)
         # TODO: check if format can be used for vector data?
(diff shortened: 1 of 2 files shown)


openeo_driver/datacube.py
```diff
@@ -587,7 +587,7 @@ def write_assets(
             log.warning(
                 f"Exporting vector cube {self} to legacy, non-standard JSON format"
             )
-            return self.to_legacy_save_result().write_assets(directory)
+            return self.to_legacy_save_result().write_assets(directory / "suffix_to_remove")
```
soxofaan, line 590: This looks like making the situation worse instead of better.

Isn't it cleaner to prevent wrongly removing suffixes instead of adding artificial suffixes under blind assumption that they are removed elsewhere

EmileSonneveld: The last part of the path is removed in many places through the code. It would be nice to get rid of at some point, especially the "/out" suffix is very obscure.

I made a new commit keeping the original argument of the first `write_assets` method, so it can explicitly be send through the second `write_assets` without needing to invent a suffix

PR #356, opened 2025-01-24
disallow Points

bossie:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/996


openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -1460,13 +1460,22 @@ def filter_labels(args: ProcessArgs, env: EvalEnv) -> DriverDataCube:
 
 def _extract_bbox_extent(args: dict, field="extent", process_id="filter_bbox", handle_geojson=False) -> dict:
     extent = extract_arg(args, name=field, process_id=process_id)
+    # TODO #114: support vector cube
     if handle_geojson and extent.get("type") in [
         "Polygon",
         "MultiPolygon",
+        "GeometryCollection",  # TODO #71 #114: deprecate GeometryCollection
         "GeometryCollection",
```
jdries, line 1468: Duplicated entry?
```suggestion
```

jdries (changes requested):
minor remark, seems fine otherwise

openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -1483,6 +1491,22 @@ def _extract_bbox_extent(args: dict, field="extent", process_id="filter_bbox", h
...
 
+def _contains_polygon(geojson: dict) -> bool:
+    if geojson["type"] in ["Polygon", "MultiPolygon"]:
+        return True
+
+    if geojson["type"] == "Feature":
+        return _contains_polygon(geojson["geometry"])
+
+    if geojson["type"] == "FeatureCollection":
+        return any(_contains_polygon(feature) for feature in geojson["features"])
```
soxofaan, line 1502: given https://github.com/Open-EO/openeo-processes/issues/527 / https://github.com/Open-EO/openeo-processes/pull/528 I think that should be `all` instead of `any`

(and the function should be named `_contains_only_polygons`)

openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -1483,6 +1491,22 @@ def _extract_bbox_extent(args: dict, field="extent", process_id="filter_bbox", h
...
+
+    if geojson["type"] == "GeometryCollection":
```
soxofaan, line 1504: I would not add support for this deprecated type, or is it necessary because we still coerce to GeometryCollection in several other places unfortunately?

Issue #358, opened 2025-01-30
global extent calculation prioritizes weak_spatial_extent over spatial_extent


JeroenVerstraelen:
https://github.com/Open-EO/openeo-python-driver/blob/master/openeo_driver/ProcessGraphDeserializer.py#L635

The order should be inverted, similar to how it is done when calculating the spatial_extent during the dry_run:
https://github.com/Open-EO/openeo-python-driver/blob/4e181db16b4eb2b3a5c2969b7c829e7825f02004/openeo_driver/dry_run.py#L483

PR #359, opened 2025-01-30
fix extent order when calculating global_extent #358

JeroenVerstraelen:


jdries:
looks good, but there's a unit test failure

jdries:
@JeroenVerstraelen can you fix the test so this can be merged?

PR #364, opened 2025-01-31
do not run top level udf in validation mode

jdries:


openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -1591,6 +1591,9 @@ def run_udf(args: dict, env: EvalEnv):
         # This way a weak_spatial_extent can be calculated from the UDF's output.
         return data.run_udf()
 
+    if env.get("validation", False):
+        raise FeatureUnsupportedException("run_udf is not supported in validation mode.")
```
soxofaan, line 1595: Validation is enabled by default in the python client since https://github.com/Open-EO/openeo-python-client/issues/404 so throwing an exception here will  trigger  warnings for (toplevel) UDF users :worried: 
Shouldn't we continue validation in best-effort style instead?

jdries: What would 'best-effort' look like? Is there something sensible we can return here?
That udf warning would be a minor inconvenience for a minority of users afaik.

soxofaan: assuming most cases of top-level UDFs are with vector cubes, we could return an empty vector cube maybe?

jdries: problem is that run_udf is also used in cases without vector cube, so those will also fail
Given that the current problem is a lot worse then the potential negative impact of confusing validation, I'll have to merge it for now.

Issue #322, opened 2024-11-05
Improve forward/backward compatibility of OpenEoBackendConfig


soxofaan:
Current `OpenEoBackendConfig` system is quite strict at the moment: 
passing unknown config values to the constructor cause an "unexpected keyword argument" exception. This is fine as default behavior (e.g. to catch config value typo's as early as possible), but can cause problems in practice. For example, sometimes we have to run older images (that don't know yet about config field X) with newer configs (that define config X). It's not ideal that this fails as hard as it currently does.

It should be possible to skip  unknown config values (and trigger warning/error logs instead of failing hard)

soxofaan:
addressed by PR #371 in 34594406cb20e458b4a4f154bcc2ee2c788e8573

PR #371, opened 2025-02-21
Issue322 config strictness

soxofaan:
for #322


openeo_driver/config/config.py
```diff
@@ -1,26 +1,24 @@
...
 from openeo_driver.urlsigning import UrlSigner
 from openeo_driver.users.oidc import OidcProvider
 from openeo_driver.workspace import Workspace
 
+__all__ = ["OpenEoBackendConfig", "openeo_config_class", "ConfigException", "check_config_definition"]
 
-class ConfigException(ValueError):
-    pass
 
+_log = logging.getLogger(__name__)
```
soxofaan, line 18: not used after all, so remove again

Issue #382, opened 2025-03-14
Clarify "version" field in EvalEnv


soxofaan:
Stumbled on this while wresting with https://github.com/Open-EO/openeo-python-driver/issues/376:

https://github.com/Open-EO/openeo-python-driver/blob/eac1f24530656ebec9e4af74d06304d2dc7e5ea1/openeo_driver/views.py#L678-L681

"version"  is pretty vague and hard to grep efficiently.

To be more clear: rename to "openeo_api_version"

PR #383, opened 2025-03-14
Issue382 openeo api version

soxofaan:
draft PR for #382


openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -301,11 +316,13 @@ def get_process_registry(self, api_version: Union[str, ComparableVersion]) -> Pr
             self._registry_cache[spec] = registry
         return self._registry_cache[spec]
 
-    def get_basic_env(self, api_version=None) -> EvalEnv:
+    def get_basic_env(self, api_version: str = OPENEO_API_VERSION_DEFAULT) -> EvalEnv:
```
soxofaan, line 305: changing  the default apparently breaks some tests, it's probably cleaner to do the default change in a separate commit

openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -437,8 +454,11 @@ def evaluate(
     """
 
     if "version" not in env:
-        _log.warning("No version in `evaluate()` env. Blindly assuming 1.0.0.")
-        env = env.push({"version": "1.0.0"})
+        # TODO #382 Deprecated field "version", use "openeo_api_version" instead
+        env = env.push({"version": OPENEO_API_VERSION_DEFAULT})
```
soxofaan, line 444: changing  the default apparently breaks some tests, it's probably cleaner to do the default change in a separate commit

PR #387, opened 2025-03-26
feature: asset_urls

pvbouwel:
Group the logic where download urls are created for assets.

The default implementation is meant to be a backwards compatible refactor.

This change does allow backend implementations to implement custom behavior such that retrieval of the actual assets does not need to go via the driver.

Notes:

- Did not go for the option to have a base implementation without signer because then it is not really possible to add this change in a backwards compatible way because consumers that use signing now must then add the first override that supports signing so signing checking logic is part of the most basic implementation (with the downside of the ugly local import)
- Test coverage should ensure backwards compatibility as all original tests still exists. With parameterize there are additional test cases to show case that an alternative AssetUrl implementation could help generate URLs against a different endpoint/service where possible and still fallback to default behavior whenever it is not possible.

### DIFF
--- a/openeo_driver/asset_urls.py
+++ b/openeo_driver/asset_urls.py
@@ -1,5 +1,7 @@
 import logging
 from openeo_driver.users import user_id_b64_encode
+import openeo_driver.config
+
 import flask
 _log = logging.getLogger(__name__)
 
@@ -19,9 +21,7 @@ class AssetUrl:
         If an url_signer is defined these urls will be signed.
 
         """
-        from openeo_driver import backend  # otherwise circular import
-
-        signer = backend.get_backend_config().url_signer
+        signer = openeo_driver.config.get_backend_config().url_signer
         if signer:
             expires = signer.get_expires()
             secure_key = signer.sign_job_asset(
--- a/openeo_driver/asset_urls.py
+++ b/openeo_driver/asset_urls.py
@@ -1,0 +1,51 @@
+import logging
+from typing import Optional
+
+from openeo_driver.users import user_id_b64_encode
+from openeo_driver.urlsigning import UrlSigner
+from functools import lru_cache
+import flask
+_log = logging.getLogger(__name__)
+
+
+class AssetUrl:
+    """
+    This class helps to convert internal storage locations to asset urls
+
+
+    If assets are not stored on local storage it is possible to come up with other implementations that extend this
+    class and go straight to the external storage. If the driver always has access to the artifacts then it is
+     recommended fallback to the default implementation by calling the method from its base class.
+    """
+    def get(self, asset_metadata: dict, asset_name: str, job_id: str, user_id: str) -> str:
+        """
+        The default implementation will create urls that go to the driver application.
+        If an url_signer is defined these urls will be signed.
+
+        """
+        signer = self._get_signer()
+        if signer:
+            expires = signer.get_expires()
+            secure_key = signer.sign_job_asset(
+                job_id=job_id, user_id=user_id, filename=asset_name, expires=expires
+            )
+            user_base64 = user_id_b64_encode(user_id)
+            return flask.url_for(
+                '.download_job_result_signed',
+                job_id=job_id, user_base64=user_base64, filename=asset_name, expires=expires, secure_key=secure_key,
+                _external=True
+            )
+        else:
+            return flask.url_for('.download_job_result', job_id=job_id, filename=asset_name, _external=True)
+
+    @lru_cache()
+    def _get_signer(self) -> Optional[UrlSigner]:
+        """
+        A helper to get a signer from config.
+
+        It uses local imports because config does contain a asset_url object so if we have an import in this package
+        it results in a circular import.
+        """
+        from openeo_driver.backend import get_backend_config
+
+        return get_backend_config().url_signer
--- a/openeo_driver/asset_urls.py
+++ b/openeo_driver/asset_urls.py
@@ -38,7 +38,6 @@ class AssetUrl:
         else:
             return flask.url_for('.download_job_result', job_id=job_id, filename=asset_name, _external=True)
 
-    @lru_cache()
     def _get_signer(self) -> Optional[UrlSigner]:
         """
         A helper to get a signer from config.
--- a/openeo_driver/asset_urls.py
+++ b/openeo_driver/asset_urls.py
@@ -19,7 +19,7 @@ class AssetUrl:
         If an url_signer is defined these urls will be signed.
 
         """
-        from openeo_driver import backend  # otherwise cicurlar import
+        from openeo_driver import backend  # otherwise circular import
 
         signer = backend.get_backend_config().url_signer
         if signer:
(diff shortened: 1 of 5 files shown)


pvbouwel:
@soxofaan  does this look ok to you? It would be a pre-req for my changes in geopyspark driver

openeo_driver/asset_urls.py
```diff
@@ -0,0 +1,50 @@
...
+class AssetUrl:
+    """
+    This class helps to convert internal storage locations to asset urls
+
+
+    If assets are not stored on local storage it is possible to come up with other implementations that extend this
+    class and go straight to the external storage. If the driver always has access to the artifacts then it is
+     recommended fallback to the default implementation by calling the method from its base class.
+    """
+    def get(self, asset_metadata: dict, asset_name: str, job_id: str, user_id: str) -> str:
```
soxofaan, line 20: I'd give this method a less generic name, e.g. `build_url`.
I've noticed that code assist tools (e.g. like in Pycharm) can go haywire when you ask to "find usage" of a generically named method (because typing is optional in python, so you're overloaded with false positives of all cases where the type can not be inferred). Picking less generic method names reduces that problem (and makes things more self-documenting as side-effect)

tests/test_views.py
```diff
@@ -115,6 +116,16 @@ def api120(client) -> ApiTester:
     return ApiTester(api_version="1.2", client=client, data_root=TEST_DATA_ROOT)
 
 
+class TestDirectS3Assets(AssetUrl):
```
soxofaan, line 119: Careful with using "Test" as class name prefix here: in pytest this has a specific meaning: it's a vehicle to group/organize test functions.
In this case you are building a stand-in/dummy, so I would go for something like "DummyS3Assets", or "CustomS3Assets", ...

tests/test_views.py
```diff
@@ -115,6 +116,16 @@ def api120(client) -> ApiTester:
...
+    S3_ENDPOINT = "https://s3.oeo.net"
```
soxofaan, line 120: unfortunately we haven't migrated all tests yet, so there are still a lot of bad examples everywhere, but for newly added unit tests, I'd work from the start with TLDs that are designed for test purposes: like `s3.test`, `s3.oeo.test`, `s3.example.com`, ...

tests/test_views.py
```diff
@@ -3018,12 +3034,22 @@ def test_get_job_result_item(self, flask_app, api110, backend_config_overrides):
...
+        ("timeseries.csv", "text/csv", {"asset_url": TestDirectS3Assets()}),
+        ("timeseries.csv", "text/csv", {}),
+        ("timeseries.parquet", "application/parquet; profile=geo", {"asset_url": TestDirectS3Assets()}),
+        ("timeseries.parquet", "application/parquet; profile=geo", {}),
     ])
-    def test_get_vector_cube_job_result_item(self, flask_app, api110, vector_item_id, vector_asset_media_type):
+    def test_get_vector_cube_job_result_item(self, flask_app, api110, vector_item_id, vector_asset_media_type, backend_config_overrides):
         vector_asset_filename = vector_item_id
+        if backend_config_overrides.get("asset_url") is None:
+            expected_asset_href = {
```
soxofaan, line 3046: I think it would be cleaner to include the expected url in the parameterization, instead of  building it here during the test. While that's anti-DRY, I think that's actually better in testing as it makes things more explicit and less magic.

openeo_driver/asset_urls.py
```diff
@@ -0,0 +1,50 @@
+import logging
+from typing import Optional
+
+from openeo_driver.users import user_id_b64_encode
+from openeo_driver.urlsigning import UrlSigner
+from functools import lru_cache
```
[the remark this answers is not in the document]
pvbouwel: Good catch, leftover from cleanup

openeo_driver/asset_urls.py
```diff
@@ -0,0 +1,50 @@
...
+            user_base64 = user_id_b64_encode(user_id)
+            return flask.url_for(
+                '.download_job_result_signed',
+                job_id=job_id, user_base64=user_base64, filename=asset_name, expires=expires, secure_key=secure_key,
+                _external=True
+            )
+        else:
+            return flask.url_for('.download_job_result', job_id=job_id, filename=asset_name, _external=True)
+
+    def _get_signer(self) -> Optional[UrlSigner]:
```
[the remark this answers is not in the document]
pvbouwel: This trick still results in a circular import:
```
ImportError while loading conftest '/home/peter/git/openeo-python-driver/tests/conftest.py'.
tests/conftest.py:13: in <module>
    import openeo_driver.config.load
openeo_driver/config/__init__.py:1: in <module>
    from openeo_driver.config.config import OpenEoBackendConfig, ConfigException
openeo_driver/config/config.py:17: in <module>
    from openeo_driver.asset_urls import AssetUrl
openeo_driver/asset_urls.py:5: in <module>
    import openeo_driver.backend
openeo_driver/backend.py:29: in <module>
    from openeo_driver.config import OpenEoBackendConfig, get_backend_config
E   ImportError: cannot import name 'OpenEoBackendConfig' from partially initialized module 'openeo_driver.config' (most likely due to a circular import) (/home/peter/git/openeo-python-driver/openeo_driver/config/__init__.py)
```

But I did eliminate the extra _get_signer method and just have the local import.

pvbouwel:
@soxofaan I addressed your feedback. The circular import did remain so still have a local import but all the other items I think were implemented as recommended (and the generic `get` method name and IDE confusion became very clear when I used pycharm to do that refactor :) ).

openeo_driver/asset_urls.py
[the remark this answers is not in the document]
soxofaan: I managed to break the cycle:
you do an import of `openeo_driver.backend` but that is quite a heavy dependency module. 
You just need  `get_backend_config()`, which is actually defined at `openeo_driver.config`, which is a lot lighter.

So this seems to work
```
# asset_urls.py

import openeo_driver.config

...

        signer = openeo_driver.config.get_backend_config().url_signer
```

openeo_driver/asset_urls.py
```diff
@@ -0,0 +1,37 @@
...
+class AssetUrl:
+    """
+    This class helps to convert internal storage locations to asset urls
+
+
+    If assets are not stored on local storage it is possible to come up with other implementations that extend this
+    class and go straight to the external storage. If the driver always has access to the artifacts then it is
+     recommended fallback to the default implementation by calling the method from its base class.
+    """
+    def build_url(self, asset_metadata: dict, asset_name: str, job_id: str, user_id: str) -> str:
```
soxofaan, line 16: API design tip:
```suggestion
    def build_url(self, *, asset_metadata: dict, asset_name: str, job_id: str, user_id: str) -> str:
```

This forces all usage to use keyword arguments and prevents off-by-one positional arguments mistakes. 
Enforcing keyword arg usage also simplifies future API evolution

soxofaan (approved):
great, looks fine to me

pvbouwel:
Rebased to have changelog entry not have conflicts

PR #391, opened 2025-04-07
support csv output for vector cube

jdries:
trivial issue fix
https://github.com/Open-EO/openeo-geopyspark-driver/issues/1136

### DIFF
--- a/tests/test_save_result_csv.py
+++ b/tests/test_save_result_csv.py
@@ -116,8 +116,8 @@ def test_aggregate_polygon_result_inconsistent_bands(tmp_path):
         result.to_csv(tmp_path / 'timeseries_invalid.csv')
 
 def test_write_driver_vector_cube_to_csv(tmp_path):
-    vector_cube = DriverVectorCube(gpd.read_file(str(get_path("geojson/FeatureCollection02.json"))))
+    vector_cube = DriverVectorCube.from_fiona([str(get_path("geojson/FeatureCollection02.json"))],driver="geojson")
     vector_cube.write_assets(tmp_path / "dummy", format="CSV")
 
-    actual_gdf = gpd.read_file(tmp_path / "vectorcube.csv")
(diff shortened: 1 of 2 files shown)


JeroenVerstraelen (approved):
LGTM

tests/test_save_result_csv.py
```diff
@@ -111,3 +114,10 @@ def test_aggregate_polygon_result_inconsistent_bands(tmp_path):
 
     with pytest.raises(IndexError):
         result.to_csv(tmp_path / 'timeseries_invalid.csv')
+
+def test_write_driver_vector_cube_to_csv(tmp_path):
+    vector_cube = DriverVectorCube(gpd.read_file(str(get_path("geojson/FeatureCollection02.json"))))
```
soxofaan, line 119: Maybe use the `DriverVectorCube.from_fiona` factory here instead of the raw constructor to be closer to the code path of actual usage?

tests/test_save_result_csv.py
```diff
@@ -111,3 +114,10 @@ def test_aggregate_polygon_result_inconsistent_bands(tmp_path):
...
+    vector_cube.write_assets(tmp_path / "dummy", format="CSV")
+
+    actual_gdf = gpd.read_file(tmp_path / "vectorcube.csv")
+    assert actual_gdf.shape == (2, 4)
```
soxofaan, line 123: `gpd.read_file()` might be to helpful/"smart" and hide the fact that the file actually is not a valid CSV but something completely different (binary format with magic header to identify it).

I would do a raw file read and check first line for example for expected column names

soxofaan (commented):
some minor notes

tests/test_save_result_csv.py
jdries: done

tests/test_save_result_csv.py
jdries: done

PR #392, opened 2025-04-07
Issue #747 added missing Content-Range header

dsamaey:


soxofaan:
related to
- https://github.com/Open-EO/openeo-python-client/issues/747

soxofaan:
looks fine to merge to me

adding a changelog entry might be useful to properly track this in deployments

PR #393, opened 2025-04-18
return NoSuchKey error as 404 response with details

bossie:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/1149


bossie:
I thought it made sense to return a 404 response with details if a user tries to download a job result that does not exist instead of a generic error like:

```
openeo.rest.OpenEoApiError: [500] Internal: Server error: NoSuchKey('An error occurred (NoSuchKey) when calling the GetObject operation: None') (ref: r-2504010710014c57bb7383ed02888e5f)
```

but considering that OpenEO offers signed URLs that can't be tampered with, links to missing assets are a bug and a 500 response might be more appropriate.

Then again, `send_from_directory` will probably also return a 404 response (TBC) if the file does not exist so it will align the two implementations:

https://github.com/Open-EO/openeo-python-driver/blob/d38a046e796ab74d309e51a16964de3294357035/openeo_driver/views.py#L1326

bossie:
So I thought about this some more and a user will complain if he can't download his file regardless of whether it's a 404 or a 500 error; this response is also clearer and conveys the fact that retrying will not help.

openeo_driver/views.py
```diff
@@ -1316,7 +1318,7 @@ def _download_job_result(
             _log.error(f"Unsupported job result: {result!r}")
             raise InternalException("Unsupported job result")
 
-    def _stream_from_s3(s3_url, result, bytes_range: Optional[str]):
+    def _stream_from_s3(s3_url, filename, mimetype: Optional[str], bytes_range: Optional[str]):
```
soxofaan, line 1321: API design tip: when number of arguments grows, positional argument handling gets more error-prone. You can [enforce keyword argument usage](https://peps.python.org/pep-3102/) with (for example):
```suggestion
    def _stream_from_s3(s3_url, *, filename, mimetype: Optional[str], bytes_range: Optional[str]):
```

openeo_driver/views.py
```diff
@@ -1328,22 +1330,22 @@ def _stream_from_s3(s3_url, result, bytes_range: Optional[str]):
...
             resp.headers['Content-Length'] = s3_file_object['ContentLength']
             if 'ContentRange' in s3_file_object:
                 resp.headers['Content-Range'] = s3_file_object['ContentRange']
             return resp
-        except s3_instance.exceptions.NoSuchKey:
-            _log.exception(f"No such key: s3://{bucket}/{key}")
-            raise
+        except s3_instance.exceptions.NoSuchKey as e:
+            _log.exception(f"Not found: {s3_url}")
+            raise OpenEOApiException(code="NotFound", status_code=404, message=f"Not found: {filename}") from e
```
soxofaan, line 1342: FYI: there is a `NotFoundException` in `openeo_driver.errors` that already takes care of code and status code

soxofaan (commented):
just some minor notes

soxofaan:
> considering that OpenEO offers signed URLs that can't be tampered with, links to missing assets are a bug

note that there are also non-signed variants for asset/result URLs (in which case auth headers are used to check access), so "missing assets" can still be because of a client error instead of server error

PR #407, opened 2025-05-19
s3 reusable code in python driver

pvbouwel:
[https://github.com/Open-EO/openeo-geopyspark-driver/issues/1195]

In order to allign S3 usage provide the integration in the Python driver.


tests/test_views.py
```diff
@@ -2895,7 +2901,7 @@ def test_download_result_vectorcube(self, api, tmp_path):
         assert "features" in resp.json.keys()
         assert resp.headers["Content-Type"] == "application/geo+json"
 
-    def test_download_result_with_s3_object_storage(self, api, mock_s3_resource):
+    def test_download_result_with_s3_object_storage(self, api, mock_s3_resource, swift_url):
```
pvbouwel, line 2904: These are the only changes to existing tests. This change is only needed to pass the tests but it is not a breaking change from a deployment perspective given that the SWIFT_URL had to be defined for old deploys otherwise it would still result in runtime errors (but that part is not covered by unittests due to its mocking behavior. In the new code a more appropriate error is returned and this is not mocked so hence it is now needed as part of the test fixture).

openeo_driver/config/config.py
```diff
@@ -88,3 +89,17 @@ class OpenEoBackendConfig(_ConfigBase):
 
     "Experimental: simple job progress fallback estimation. Specify average batch job completion time (wall clock) in seconds."
     simple_job_progress_estimation: Optional[float] = None
+
+    "When using S3 compatible object storage the supported regions and endpoints should be defined here"
+    s3_provider_config: ProvidersCfg = ProvidersCfg(
```
soxofaan, line 94: I'm not sure if it's useful to require  `s3_provider_config` to be a `ProvidersCfg`, it looks like it can just be a dict. 
That would also eliminate the dependency of this config module on `openeo_driver.integrations.s3.provider_cfg`  which might drag in additional dependencies that are only optional (e.g. s3 functionality is still optional, e.g. not used in openeo-aggregator)

soxofaan, line 94: I'm not sure if it's useful to require  `s3_provider_config` to be a `ProvidersCfg`, it looks like it can just be a dict. 
That would also eliminate the dependency of this config module on `openeo_driver.integrations.s3.provider_cfg`  which might drag in additional dependencies that are only optional (e.g. s3 functionality is still optional, e.g. not used in openeo-aggregator)

soxofaan: Or another design pattern I used for Calrissian config:

[openeogeotrellis.config.integrations.calrissian_config.CalrissianConfig](https://github.com/Open-EO/openeo-geopyspark-driver/blob/master/openeogeotrellis/config/integrations/calrissian_config.py) is a config container defined under `openeogeotrellis/config/integrations`, far away from and consequently without any of the fancy dependencies  of the the actual implementation stuff at `openeogeotrellis/integrations/calrissian`.
This eliminates al the dependency hell risk when importing CalrissianConfig from the main config module

felt like a good compromise of make a toplevel config field (`calrissian_config`) a bit more powerful than a naive dict, with low risk of feature creeping into dependency hell

openeo_driver/integrations/s3/client.py
```diff
@@ -0,0 +1,45 @@
...
+
+class S3ClientBuilder:
+    @classmethod
+    def from_region(cls, region_name: Optional[str]) -> S3Client:
+        if region_name is None:
+            _log.warning(
+                "Building S3 client without region. This should be avoided as much as possible because these operations"
+                "will be limited to a single endpoint for which the platform is configured."
+            )
+            region_name = os.environ.get("AWS_REGION", "UNKNOWN")
```
soxofaan, line 24: Does a region "UNKNOWN" ever make sense? Isn't it better to just fail instead?

openeo_driver/integrations/s3/credentials.py
```diff
@@ -0,0 +1,48 @@
+import logging
+import os
+from typing import TypedDict, Optional
+
+_log = logging.getLogger(__name__)
+
+
+class Boto3CredentialsTypeDef(TypedDict):
```
soxofaan, line 8: I would not add `TypeDef` suffix, that makes it trivial to evolve this class to real class that's more than just a type annotation.

openeo_driver/integrations/s3/credentials.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    for prefix in [f"{region_name}", f"{provider_name}", "SWIFT", "AWS"]:
+        credential_candidate = get_credential_for_prefix(prefix)
+        if credential_candidate is not None:
+            return credential_candidate
+    raise EnvironmentError(f"Cannot get credentials for {region_name}")
+
+
+def _sanitize_env_name(env_name: str) -> str:
+    return env_name.upper().replace("-", "_")
```
soxofaan, line 30: why not go full regex to be sure? Something like  (untested)

```suggestion
    re.sub("[^A-Z0-9]+", "_", env_name.upper())
```

soxofaan (commented):
some initial notes

(train is arriving :smile:  )

openeo_driver/config/config.py
pvbouwel: I like the 2nd suggestion better. I already split out the content in a separate file module but I agree that putting it separate is better to avoiding a dependency hell.

openeo_driver/integrations/s3/client.py
pvbouwel: Some S3 implementations do not check the region provided as long as you speak with the endpoint. So an AWS_REGION of UNKNOWN would actually work given that SWIFT_URL is set correctly (if that one is not set correctly we would FAIL)

openeo_driver/config/config.py
pvbouwel: Took the separate config location strategy.

openeo_driver/integrations/s3/presigned_url.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    try:
+        response = client.generate_presigned_url(
+            "get_object",
+            Params={"Bucket": bucket_name, "Key": object_name},
+            ExpiresIn=expiration,
+        )
+    except ClientError as e:
+        logging.info(f"Failed to create presigned url: {e}, continuing without.")
+        return None
```
soxofaan, line 28: is this `create_presigned_url` being used in the openeo-python-client itself?

and is returning None on failure a good default behavior?

tests/integrations/test_s3.py
```diff
@@ -0,0 +1,174 @@
...
+        pytest.param(
+            "waw4-1",
+            "https://s3.waw4-1.cloudferro.com",
+        ),
+        pytest.param("eu-nl", "https://obs.eu-nl.otc.t-systems.com"),
+        pytest.param("EU-NL", "https://obs.EU-NL.otc.t-systems.com"),
+        pytest.param("eu-de", "https://obs.eu-de.otc.t-systems.com"),
+        pytest.param("eu-faketest-central", legacy_test_endpoint),
+    ],
+)
```
soxofaan, line 54: FYI: it's not necessary to use `pytest.param` in this case (I only use it, rarely, to override the default id in reporting).

you can just specify simple tuples, like:
```python
@pytest.mark.parametrize(
    ["region_name", "expected_endpoint"],
    [
        ("waw3-1", "https://s3.waw3-1.cloudferro.com"),
        ("waw3-2", "https://s3.waw3-2.cloudferro.com"),
...
```

And also note that you can specify the parameter names transparantly as a list of strings ` ["region_name", "expected_endpoint"]`, instead of some ~~ugly~~ adhoc CSV format `"region_name,expected_endpoint"`

soxofaan (commented):
apart from some small notes, I guess this is ready to merge

PR #410, opened 2025-05-21
Issue #1128 avoid realignment for resolutions > 20m

dsamaey:


openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -583,14 +584,17 @@ def _align_extent(extent,collection_id,env,target_resolution=None):
...
-            and extent.get('crs','') == "EPSG:4326"
-            and "extent" in x and "extent" in y
-            and (target_resolution == None or  target_resolution == collection_resolution)
+    if (
+        crs == 4326
+        and extent.get("crs", "") == "EPSG:4326"
+        and "extent" in x
+        and "extent" in y
+        and (target_resolution is None or target_resolution == collection_resolution)
+        and collection_resolution[0] <= 20
```
jdries, line 593: when crs is 4326, target_resolution will be specified in degrees, so can not compare to 20m.
I was thinking of extending the clause over here:
https://github.com/Open-EO/openeo-python-driver/blob/58bf13016c540d6ee3eb093d0c7e1f34c25ce6fc/openeo_driver/ProcessGraphDeserializer.py#L680

with something like this:
`no_resampling = "UTM zone" in CRS.from_user_input(crs).to_wkt() and target_resolution <=20`
(I didn't actually test it)

jdries (changes requested):

PR #414, opened 2025-06-23
feature: support custom parameters for presigned urls

pvbouwel:
Allow passing extra parameters starting with `x-` which can be added to presigned urls.

`x-` is used as extension values so this allows to provide extra extension values for the target of the presigned urls which OpenEO requires to inform the proxy that an URL can be used for both head and get requests.


soxofaan (approved):
looks fine to me

but maybe you want to bump the "alpha: version number of https://github.com/Open-EO/openeo-python-driver/blob/7705f4d5b356018f6b22373d5b72edfe1a0b517c/openeo_driver/_version.py#L1  (or add a changelog entry) for easier follow up of the deploy of this feature?

PR #417, opened 2025-07-10
1111 avoid relative asset path assumption

bossie:
For self-review only.


bossie (commented):
Addressed inline.

bossie:
Implemented in https://github.com/Open-EO/openeo-python-driver/pull/412 instead.

PR #422, opened 2025-09-05
convert_node: more fine-grained result caching control

soxofaan:
for Open-EO/openeo-geopyspark-driver#1331

- Introduce _ResultCachingMode with fine-grained result caching options 
- make it more clear that original behavior is highly geopyspark specific


jdries (approved):

jdries:
looks good! 
The caching mechanism for load_collection is more specific, allowing multiple return values for a single load_collection node, which is needed.

EmileSonneveld:
I just checked the original graph with `job_options = {"node_caching": "full"}` and it became way faster to reach `logger.info("Evaluated process graph` (without this setting, it does not reach this point after 15 minutes). Seems safe to me

PR #423, opened 2025-09-08
support signed links to auxiliary files in batch job result STAC items

bossie:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/1278


bossie:
Adds links with `"_expose_internal: True"` to batch job result items.

PR #430, opened 2025-10-21
issue 1111 set list_job_results list items with stac1.1

ElienVandermaesenVITO:
follow up on https://github.com/Open-EO/openeo-python-driver/pull/411


openeo_driver/views.py
```diff
@@ -1187,10 +1188,16 @@ def job_result_item_url(item_id, is11 = False) -> str:
 
 
             if len(result_metadata.items) > 0 :
```
bossie:
```suggestion
            if result_metadata.items:
```

I would take the opportunity to replace this as it it shorter, safer and expresses the intent better.

tests/test_views.py
```diff
@@ -2930,6 +2930,9 @@ def test_get_job_results_from_stac_1_1_items(self, api110, backend_config_overri
 
             resp = api110.get(f"/jobs/{job_id}/results", headers=self.AUTH_HEADER).assert_status_code(200)
 
+            assert "assets" in resp.json
```
bossie: Test the actual value of this `assets` property.

EmileSonneveld:
@codex review

PR #447, opened 2025-12-23
Initial implementation of CWL support in run_udf. 

EmileSonneveld:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/1476
@codex review

### DIFF
--- a/openeo_driver/ProcessGraphDeserializer.py
+++ b/openeo_driver/ProcessGraphDeserializer.py
@@ -1785,7 +1785,7 @@ def run_udf(args: ProcessArgs, env: EvalEnv):
         return env.backend_implementation.run_cwl(
             data=data,
             env=env,
-            udf=udf,
+            cwl=udf,
             context=context,
         )
 


openeo_driver/ProcessGraphDeserializer.py
```diff
@@ -1812,6 +1812,12 @@ def run_udf(args: ProcessArgs, env: EvalEnv):
             structured_data_list=[openeo.udf.StructuredData(description="Data list", data=data, type="list")],
             user_context=context
         )
+    elif runtime.lower() == "CWL-Calrissian".lower():
```
jdries: better runtime name: EOAP or else perhaps EOAP-CWL

openeo_driver/backend.py
```diff
@@ -949,6 +949,16 @@ def load_ml_model(self, job_id: str) -> DriverMlModel:
     def vector_to_raster(self, input_vector_cube: DriverVectorCube, target: DriverDataCube) -> DriverDataCube:
         raise NotImplementedError
 
+    def run_cwl(
+        self,
+        env: EvalEnv,
+        cwl_url: str,
```
soxofaan: technically, the `udf` arg of `run_udf` openeo process can be:
- file path
- url
- code bloc

so I would make this `cwl_url` more generic: e.g. just  `cwl` or maybe even `udf`

EmileSonneveld: Renamed to cwl

Issue #451, opened 2026-01-08
STAC 1.1: invalid metadata returned


jdries:
I ran a job that produced STAC 1.1 metadata, but except for the version, most metadata is still a weird combo of 1.0 and 1.1 things:

```
{
  "assets": {
    "openEO": {
      "bands": [
        {
          "name": "VV_P10"
        },
        {
          "name": "VV_P25"
        },
        {
          "name": "VV_P50"
        },
        {
          "name": "VV_P75"
        },
        {
          "name": "VV_P90"
        },
        {
          "name": "VH_P10"
        },
        {
          "name": "VH_P25"
        },
        {
          "name": "VH_P50"
        },
        {
          "name": "VH_P75"
        },
        {
          "name": "VH_P90"
        },
        {
          "name": "VHVV-RATIO_P10"
        },
        {
          "name": "VHVV-RATIO_P25"
        },
        {
          "name": "VHVV-RATIO_P50"
        },
        {
          "name": "VHVV-RATIO_P75"
        },
        {
          "name": "VHVV-RATIO_P90"
        }
      ],
      "eo:bands": [
        {
          "name": "VV_P10"
        },
        {
          "name": "VV_P25"
        },
        {
          "name": "VV_P50"
        },
        {
          "name": "VV_P75"
        },
        {
          "name": "VV_P90"
        },
        {
          "name": "VH_P10"
        },
        {
          "name": "VH_P25"
        },
        {
          "name": "VH_P50"
        },
        {
          "name": "VH_P75"
        },
        {
          "name": "VH_P90"
        },
        {
          "name": "VHVV-RATIO_P10"
        },
        {
          "name": "VHVV-RATIO_P25"
        },
        {
          "name": "VHVV-RATIO_P50"
        },
        {
          "name": "VHVV-RATIO_P75"
        },
        {
          "name": "VHVV-RATIO_P90"
        }
      ],
      "href": "https://openeo.dataspace.copernicus.eu/openeo/1.2/jobs/j-2601081205244c6bafa822295190625d/results/assets/MzJjYzdkZGItZjdlMS00YjFjLTk3OTYtZjlmZTM5Y2I4ZmVi/d603b88c97c3ca5bac4f85de07fc9e6f/openEO?expires=1768487348",
      "proj:epsg": 32629,
      "raster:bands": [
        {
          "name": "VV_P10",
          "statistics": {
            "maximum": 843.55926513672,
            "mean": 0.11038660176015,
[...]

PR #453, opened 2026-01-12
include "bands" iff STAC 1.1

bossie:
#451


jdries:
For my understanding: why do we still do this complex item building inside the web app, when the batch already generates full items with correct assets?
Isn't it better to fully push that responsibility to batch jobs, and let the webapp just return the metadata 'as-is' without extra wrangling?

bossie:
> For my understanding: why do we still do this complex item building inside the web app, when the batch already generates full items with correct assets?
Isn't it better to fully push that responsibility to batch jobs, and let the webapp just return the metadata 'as-is' without extra wrangling?

A couple of reasons but the main one is: it's the least intrusive fix that avoids having to touch code in different projects so I can get it done in time.

Others:
- there's pre-1.1 STAC that I don't want to break;
- there's an issue with netCDF statistics that @ElienVandermaesenVITO is looking into at the same time and I don't want these things interfering;
- there is still some mapping to be done in the python-driver regardless;
- https://github.com/Open-EO/openeo-geopyspark-driver/issues/440 can still be addressed to clean things up.

PR #481, opened 2026-04-10
Match env parameter name for query_stac

Hneuschmidt:
Fixes a typo from #480


EmileSonneveld:
Should be on CDSE staging now

PR #495, opened 2026-04-29
provide pg/tracer to log added value from

### LINKED ISSUE #1436: ETL API: report added value for sync requests
Requested by @JanssenBrm: like batch jobs, sync requests should report added value in addition to resource usage.

The information that we currently lack in at [this point in the code](https://github.com/Open-EO/openeo-geopyspark-driver/blob/03243eb2b0001693e69efd96dff68d2b572159c3/openeogeotrellis/backend.py#L1279-L1340) is:
- area
- processes called

bossie:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/1436


openeo_driver/backend.py
```diff
@@ -786,7 +787,7 @@ class Processing(MicroService):
     def get_process_registry(self, api_version: Union[str, ComparableVersion]) -> ProcessRegistry:
         raise NotImplementedError
 
-    def evaluate(self, process_graph: dict, env: EvalEnv = None):
+    def evaluate(self, process_graph: dict, env: EvalEnv = None, do_dry_run: Union[bool, DryRunDataTracer] = True):
```
soxofaan: Hmm, not a big fan of this change: this makes the dry-run aspect leak into the generic `Processing.evaluate()` API, which is not intended to be concerned with dry and wet run aspects.

For example: this will break the aggregator, which also implements this API, but has no dry-run concept

soxofaan: Isn't it possible to pass the tracer trough the existing EvalEnv instead?

bossie: Something like this, roughly?

Open to naming suggestions.

PR #508, opened 2026-06-03
feature: Bounded TTL Caching

pvbouwel:
Make sure the TtlCache always has an upper bound on the items it can contain.

Breaking changes:
- No longer support per item TTL warning is emitted upon usage.
- Boundary semantics are different at the exact boundary the element is expired. We don't rely on second accuracy and the new behavior is safer.

### DIFF
--- a/openeo_driver/util/caching.py
+++ b/openeo_driver/util/caching.py
@@ -1,7 +1,11 @@
 import functools
 import logging
+import threading
 import time
-from typing import Union, Tuple, Any, Callable, Dict, Optional
+import warnings
+from typing import Union, Tuple, Any, Callable, Optional
+
+import cachetools
 
 _log = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


openeo_driver/util/caching.py
```diff
@@ -11,62 +15,88 @@
...
+
+    Cache interactions are thread-safe. The lock is intentionally *not* held while
+    a cache-miss callback is executing, so slow callbacks do not block other readers.
     """
 
     def __init__(
-        self, default_ttl: float = 60, _clock: Callable[[], float] = time.time
+        self,
+        default_ttl: float = 60,
+        *,
```
soxofaan, line 29: also make default_ttl a keyword-only argument

openeo_driver/util/caching.py
```diff
@@ -11,62 +15,88 @@
...
+        max_size: int = 1000,
+        _clock: Callable[[], float] = time.time,
```
soxofaan, line 31: Given that the tests now use time_machine to time mocking, is it still necessary to have this `_clock` attribute (which is only there for testability)?

openeo_driver/util/caching.py
soxofaan: And it also doesn't make sense anymore to call it `default_ttl`: should just be `ttl`

soxofaan (commented):
the complexity of this PR is a bit bigger than I expected, especially because it introduces some breaking of the original API

maybe it's just easier to leave the original TtlCache as is, and just introduce a new BoundedTtlCache class?

pvbouwel:
I wouldn't mind a BoundedTtlCache then it is easier to address comments like signature changes but I'd still deprecate the old class because otherwise if issues are encountered down the road there are 2 implementations to maintain and the replacement protects more against mistakes.

Can as well be 2 separate steps. Add the class and "switch usage and deprecate old class" that way deployment would be staightforward and rollback would be easy.

tests/users/test_auth.py
```diff
@@ -361,7 +363,7 @@ def userinfo(request, context):
         assert internal_auth_data["access_token"] == oidc_access_token
 
 
-def test_bearer_auth_oidc_caching(app, requests_mock, oidc_provider):
+def test_bearer_auth_oidc_caching(app, requests_mock, oidc_provider, simple_time):
```
soxofaan, line 366: I think using the "fast_sleep" fixture here would be more to the point and more self-explanatory

tests/users/test_auth.py
```diff
@@ -370,58 +372,48 @@ def userinfo(request, context):
...
 
-        set_time(10000)
+        time.sleep(10000)
         resp = client.get("/personal/hello", headers=headers)
         assert (resp.status_code, resp.data) == (200, b"hello oidcuser")
         assert userinfo.call_count == 3
 
 
-def test_userinfo_url_caching(app, requests_mock, oidc_provider):
+def test_userinfo_url_caching(app, requests_mock, oidc_provider, simple_time):
```
soxofaan, line 398: same note about fast_sleep

soxofaan:
another way to simplify this task is maybe first to eliminate the per-`set()` ttl feature and only have a instance wide ttl, and do this really "breaking" (without warning), because this feature is only used internally, so we can migrate in the same commit.

And then the integration with cachetool will be a lot cleaner of a PR

pvbouwel:
If class is to be separated that would be https://github.com/Open-EO/openeo-python-driver/pull/509

tests/users/test_auth.py
pvbouwel: I did is to be consistent with OpenEO Python client. The problem with fast_sleep is that it does not set a reference time. So simple_time does `time_machine.move_to(1000, tick=False)` and then allows fast_sleep. If you just use fixture `fast_sleep` still need to initialize time_machine prior to using

pvbouwel:
@soxofaan  so the swap PR would be as simple as https://github.com/Open-EO/openeo-python-driver/pull/510 which is on top of #509 

This way it is also a cleaner PR there will just be a follow-up cleanup PR (removal of the old class) which will also be simple.

soxofaan:
:+1: 
merged #509 and #510, so I guess this one can be closed

PR #512, opened 2026-06-11
Test custom job results

EmileSonneveld:
@copilot review


EmileSonneveld:
@copilot review

openeo_driver/views.py
```diff
@@ -1850,7 +1850,7 @@ def _asset_object(
         if filename.endswith(".model"):
             # Machine learning models.
             return result_dict
-        bands = asset_metadata.get("bands")
+        bands = asset_metadata.get("bands") or asset_metadata.get("eo:bands")
```
EmileSonneveld: @soxofaan An issue for CWL here was that the results can be stac1.0, while the code assumed the was 1.1. 
Would wrapping this `or` with `TREAT_JOB_RESULTS_V100_LIKE_V110` make more sense?

EmileSonneveld: For example with 
```python
        bands = asset_metadata.get("bands")
        TREAT_JOB_RESULTS_V100_LIKE_V110 = smart_bool(os.environ.get("TREAT_JOB_RESULTS_V100_LIKE_V110", "0"))
        if not bands and stac11 and TREAT_JOB_RESULTS_V100_LIKE_V110:
            bands = asset_metadata.get("eo:bands")
```

openeo_driver/util/stac_utils.py
```diff
@@ -106,9 +106,11 @@ def recurse(item_path: Union[str, Path]) -> list:
                 href = robust_urljoin(item_path, href)
 
                 if "rel" in link and (link["rel"] == "child" or link["rel"] == "item"):
-                    all_files.extend(get_files_from_stac_catalog(href, include_metadata))
-                else:
+                    all_files.extend(recurse(href))
+                elif ("rel" not in link) or (link["rel"] != "collection"):
```
soxofaan: this is bit hard to read because of multiple negations

but regardless, the deny-list pattern (accept everything, except rel=collection) looks dangerous, as there might be a lot of other rel types you want to skip. Isn't it better to use an allow list?

openeo_driver/views.py
soxofaan: as far as I understand, `TREAT_JOB_RESULTS_V100_LIKE_V110` is a temporary toggle/hack, which I would avoid here

I would just use `bands = asset_metadata.get("bands") or asset_metadata.get("eo:bands")`

There is a possible problem however that "bands" and "eo:bands" have subtly different sub-structure: e.g. in "bands" you could find "center_wavelength", while in "eo:bands" you would just have "center_wavelength" (without prefix). But I guess the most (only?) important field is just "name", which is the same in both.

tests/data/example_stac_catalog_11/openEO_2023-06-01Z.tif
soxofaan, line 1: I'm not a fan of adding concrete TIFF files to the openeo-python-driver git repo. As far as I know, there is no functionality in the python driver that actually reads these files, so adding binary files feels only comes with costs and zero benefits. I think you don't need the TIFF files, and it's fine to have dangling hrefs in the metadata in this case.

tests/data/example_stac_catalog_11/item-2023-06-06.json
soxofaan, line 1: is it intentional to have these obscure file names for the items? this could just be something like "item01.json", right?

tests/data/job_metadata_from_cwl.json
soxofaan, line 1: what is this document exactly? It doens't have an id or type as far as I see

tests/data/job_metadata_from_cwl.json
```diff
@@ -0,0 +1,308 @@
+{
+  "geometry": null,
+  "bbox": null,
+  "area": null,
+  "start_datetime": "2023-06-01T00:00:00Z",
+  "end_datetime": "2023-06-06T00:00:00Z",
+  "links": [
+    {
+      "href": ".../collection.json",
```
soxofaan: Is it intentional to have leading tripple dots here (and in some other places)?

tests/data/example_stac_catalog_11/item-2023-06-06.json
EmileSonneveld: It is the direct output of a openeo job with stac1.1. I logged something for it here: https://github.com/Open-EO/openeo-geopyspark-driver/issues/1709#issuecomment-4689256741
I'll already give better filenames here, because the filenames are probably not valid in windows

openeo_driver/util/stac_utils.py
EmileSonneveld: simplified the logic

tests/data/job_metadata_from_cwl.json
EmileSonneveld: Changed this and other examples to `.../non/existing/example/path/collection.json`

tests/data/example_stac_catalog_11/openEO_2023-06-01Z.tif
EmileSonneveld: removed tiff files. Tests still work.

tests/data/job_metadata_from_cwl.json
EmileSonneveld: It is the batch job metadata that gets stored in the results folder. It is used to create stac content with that is returned to the client.

openeo_driver/views.py
EmileSonneveld: Removed the conditional, and now only keeps name property
`bands = [{"name": b["name"]} for b in eo_bands]`

tests/data/example_stac_catalog_11/item-2023-06-06.json
EmileSonneveld: MR to avoid commiting file names that do not work on windows: https://github.com/Open-EO/openeo-python-driver/pull/514

openeo_driver/util/stac_utils.py
EmileSonneveld: logic simplified with whitelisting: `elif "roles" in link and ("data" in link["roles"]):`

tests/test_views.py
```diff
@@ -2314,6 +2408,22 @@ def test_get_job_results_110(self, api110):
...
+            assert "j-job_metadata_from_cwl" in resp_txt
+            # assert "bands" in resp_txt  # TODO?
+            pystac.Item.from_dict(json.loads(resp_txt))
+
+    def test_get_job_results_110_from_cwl(self, api110):
+        with self._fresh_job_registry(next_job_id="j-job_metadata_from_cwl"):
+            resp = api110.get("/jobs/j-job_metadata_from_cwl/results", headers=self.AUTH_HEADER)
+            resp_txt = resp.text
+            assert "j-job_metadata_from_cwl" in resp_txt
+            assert "bands" in resp_txt
```
EmileSonneveld, line 2424: TODO: More formal "bands" check that does not match eo:bands
TODO: Also check if custom properties get trough

EmileSonneveld:
The Brokman CWL exports stac version 1.1, so this is less critical for them.
InSar exports 1.0, so this would be still needed. Unless I change the insar process.

jdries:
@EmileSonneveld 1.1 is the new default for all new features, so yes, insar may need an update

EmileSonneveld:
ok, made a ticket: https://github.com/cloudinsar/s1-workflows/issues/142
I don't know yet what version of stac BiaB exports, but it is under development and can probably be changed easily if needed.

Issue #519, opened 2026-06-25
Fix `NO2Covid.ipynb`


EmileSonneveld:
https://github.com/eu-cdse/notebook-samples/blob/main/openeo/NO2Covid.ipynb
gives: `AttributeError: 'GeoSeries' object has no attribute 'geoms'`
Found by user: https://forum.dataspace.copernicus.eu/t/openeo-sentinel-5p-no-download-fails-beginner-looking-for-guidance/5263
Probably related: https://github.com/Open-EO/openeo-python-driver/issues/158

EmileSonneveld:
Tested the notebook on CDSE dev, using SENTINEL2 instead of S5 that needs sentinel hub: `j-260706083923494cb2ea085a6fd849be`

EmileSonneveld:
This is deployed on CDSE prod and tested. `j-2608131115534027a56edd1734bf443f` `j-26081311213342af97a2b7820fe100cb`

PR #520, opened 2026-06-25
Only do geometries.geoms when the attribute is present.

EmileSonneveld:
https://github.com/Open-EO/openeo-python-driver/issues/519
@copilot review

### DIFF
--- a/openeo_driver/save_result.py
+++ b/openeo_driver/save_result.py
@@ -20,7 +20,6 @@ import typing
 from deprecated import deprecated
 from flask import send_from_directory, jsonify, Response
 import shapely.geometry
-from geopandas import GeoSeries
 from shapely.geometry import GeometryCollection
 from shapely.geometry.base import BaseGeometry
 import geopandas as gpd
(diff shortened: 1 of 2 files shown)


openeo_driver/save_result.py
```diff
@@ -446,7 +446,8 @@ def _create_point_timeseries_xarray(self, feature_ids, timestamps, lats, lons, a
 
     def to_netcdf(self, destination: Optional[str] = None) -> str:
         def features_ids_from_index(geometries):
-            return ["feature_%d" % i for i in range(len(geometries.geoms))]
+            geoms = geometries.geoms if hasattr(geometries, "geoms") else geometries
+            return ["feature_%d" % i for i in range(len(geoms))]
```
soxofaan: `features_ids_from_index` is already a bit obscure because of lack of type annotations. Adding this if here makes it worse I think

I would fully eliminate `features_ids_from_index` and just do the list comprehension inline in the if-else below. Something like

```python
        if isinstance(self._regions, GeometryCollection):
            ...
            feature_ids = ["feature_%d" % i for i in range(len(self._regions))]
        elif isinstance(self._regions, DriverVectorCube):
            ...
            feature_ids = (list(feature_ids) if feature_ids is not None
                           else ["feature_%d" % i for i in range(self._regions.geometry_count())])
        else:
            raise ValueError(self._regions)

```

(Note: GeometryCollection is something we should get rid of, so this would prepare more cleanly for that)

openeo_driver/save_result.py
```diff
@@ -314,8 +314,7 @@ def __init__(
         """
         super().__init__(data=timeseries)
         if not isinstance(regions, (GeometryCollection, DriverVectorCube)):
-            # TODO: raise exception instead of warning?
-            _log.warning(
+            raise ValueError(
```
EmileSonneveld, line 317: @soxofaan I did [not find this warning in kibana](https://kibana-infra.vgt.vito.be/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-2M,to:now))&_a=(columns:!(levelname,name,message),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'592a42f0-e665-11ec-8cc4-3747d5233c59',key:levelname,negate:!f,params:(query:WARNING),type:phrase),query:(match_phrase:(levelname:WARNING)))),hideChart:!f,index:'592a42f0-e665-11ec-8cc4-3747d5233c59',interval:auto,query:(language:kuery,query:'message:%20AggregatePolygonResult*'),sort:!(!('@timestamp',desc)))). Safe to convert it to Exception?

openeo_driver/save_result.py
EmileSonneveld: I used ai first. Could have been better. I used your suggestion now. That works fine

openeo_driver/save_result.py
soxofaan: yes makes sense

EmileSonneveld: Apparently it does get triggered in Jenkins. This now errors in CI when regions is a path to a geojson file. https://jenkins.vgt.vito.be/job/openEO/job/openeo-geopyspark-driver/job/master/3750/consoleText

PR #522, opened 2026-06-29
Fix for DelayedVector that are not geojson format.

EmileSonneveld:
Found while fixing typings with mypy
Related https://github.com/Open-EO/openeo-geopyspark-driver/issues/793

### DIFF
--- a/openeo_driver/save_result.py
+++ b/openeo_driver/save_result.py
@@ -26,6 +26,8 @@ import geopandas as gpd
 import xarray
 
 from openeo.metadata import CollectionMetadata
+from typeguard import check_argument_types
+
 from openeo_driver.datacube import DriverDataCube, DriverVectorCube, DriverMlModel
 from openeo_driver.datastructs import StacAsset
 from openeo_driver.delayed_vector import DelayedVector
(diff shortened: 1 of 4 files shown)


EmileSonneveld:
@copilot review without making changes

openeo_driver/save_result.py
```diff
@@ -1010,7 +1010,7 @@ def to_save_result(data: Any, format: Optional[str] = None, options: Optional[di
             return JSONResult(data.geojson, format="geojson", options=options)
         else:
             data = data.to_driver_vector_cube()
-    elif isinstance(data, DriverDataCube):
+    if isinstance(data, DriverDataCube):
```
EmileSonneveld, line 1015: @JeroenVerstraelen I found this while checking with mypy. This makes sure the 'data' used in the object that is returned

openeo_driver/save_result.py
```diff
@@ -26,6 +26,8 @@
 import xarray
 
 from openeo.metadata import CollectionMetadata
+from typeguard import typechecked
```
soxofaan, line 29: is this import of typeguard intended, or just temporary during development?

because typeguard was only added as test-dependency in setup.py,
while importing it here in save_result makes it  a runtime dependency

EmileSonneveld: Yes, this does not work handily. I don't know if it has a performance cost yet, so it would be risky to run it in production directly. I'll close this PR for the moment and only check typing when running in tests.

Issue #524, opened 2026-08-10
feature_id_property not used in CSV output


EmileSonneveld:
https://forum.dataspace.copernicus.eu/t/access-to-geojson-properties/5418

EmileSonneveld:
Now on production
https://openeo.dataspace.copernicus.eu/openeo/1.2/jobs/j-26081913314747d1b10bd95f8aac244b/results/YTc3ZDBjMjgtNGE3OC00YzBjLThlODUtOTMyNjQzYWQ5ZjU2/d8b00282de4da27d8228b3a5f0fed8e9?expires=1787751864

```json
{
  "process_graph": {
    "loadcollection1": {
      "process_id": "load_collection",
      "arguments": {"id": "SENTINEL2_L2A", "spatial_extent": {"west": 0, "south": 0, "east": 9, "north": 9, "crs": "EPSG:4326"}, "temporal_extent": ["2019-10-15", "2019-10-16"], "bands": ["B04"]}
    },
    "aggregatespatial1": {
      "process_id": "aggregate_spatial",
      "arguments": {
        "data": {"from_node": "loadcollection1"},
        "geometries": {
          "type": "FeatureCollection",
          "features": [
            {"type": "Feature", "properties": {"id": "f_north"}, "geometry": {"type": "Polygon", "coordinates": [[[0, 0], [5, 1], [1, 4], [0, 0]]]}},
            {"type": "Feature", "properties": {"id": "f_middle"}, "geometry": {"type": "Polygon", "coordinates": [[[6, 1], [1, 7], [9, 9], [6, 1]]]}},
            {"type": "Feature", "properties": {"id": "f_south"}, "geometry": {"type": "Polygon", "coordinates": [[[6, 1], [1, 7], [9, 9], [6, 1]]]}}
          ]
        },
        "reducer": {"process_graph": {"mean1": {"process_id": "mean", "arguments": {"data": {"from_parameter": "data"}}, "result": true}}}
      }
    },
    "saveresult1": {"process_id": "save_result", "arguments": {"data": {"from_node": "aggregatespatial1"}, "format": "CSV", "options": {"feature_id_property": "id"}}, "result": true}
  }
}
```

PR #525, opened 2026-08-10
Feature id property csv

EmileSonneveld:
Now when adding `feature_id_property` to the `save_result`, it adds a column with the feature id. Just like parquet.
@soxofaan I disabled adding this column by default, for backwards compatibility, even tough no test failed on it. Ok for you?

image
image

https://github.com/Open-EO/openeo-python-driver/issues/524

### DIFF
--- a/openeo_driver/save_result.py
+++ b/openeo_driver/save_result.py
@@ -842,10 +842,9 @@ class AggregatePolygonResultCSV(AggregatePolygonResult):
 
             if id_mapping is not None:
                 df = pd.read_csv(destination)
-                if "feature_index" in df.columns:
-                    column_name = self._feature_id_column_name()
-                    df[column_name] = df["feature_index"].map(id_mapping)
-                    df.to_csv(destination, index=False)
+                column_name = self._feature_id_column_name()
+                df[column_name] = df["feature_index"].map(id_mapping)
+                df.to_csv(destination, index=False)
 
             return destination
 
--- a/openeo_driver/save_result.py
+++ b/openeo_driver/save_result.py
@@ -781,6 +781,25 @@ class AggregatePolygonResultCSV(AggregatePolygonResult):
             return self.to_covjson()
         return self.data
 
+    def _feature_id_mapping(self) -> Optional[dict]:
+        feature_id_property = self._feature_id_column_name()
+        if not isinstance(self._regions, DriverVectorCube):
+            _log.warning(
+                f"save_result: 'feature_id_property' was specified, but regions are not a DriverVectorCube ({type(self._regions)})."
+            )
+            return None
+        values = self._regions.get_band_values(feature_id_property)
+        if values is None:
+            _log.warning(
+                f"save_result: a feature_id_property '{feature_id_property}' was specified,"
+                f" but could not find it in the vector cube: {self._regions}."
+            )
+            return None
+        return dict(enumerate(values))
+
+    def _feature_id_column_name(self) -> str:
+        return self.options.get("feature_id_property", "id")
+
     def to_csv(self, destination=None):
         csv_paths = glob.glob(self._csv_dir + "/*.csv")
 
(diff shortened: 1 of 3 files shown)


tests/test_save_result_csv.py
```diff
@@ -120,3 +122,29 @@ def test_write_driver_vector_cube_to_csv(tmp_path):
...
+    )
+    result.set_format("csv", options={"feature_id_property": "id"})
+
+    filename = result.to_csv(tmp_path / "timeseries_feature_id.csv")
+    assert filename
+    actual_df = pd.read_csv(filename)
+    assert "id" in actual_df.columns
+    assert set(actual_df["id"].dropna()) == {"f_north", "f_south"}
+    # feature_index 1 (f_middle) is not present in the source CSVs
+    assert "f_middle" not in set(actual_df["id"].dropna())
```
soxofaan: this assert is not necessary, right? 
the assert 2 lines higher already covers that?

openeo_driver/save_result.py
```diff
@@ -781,14 +781,38 @@ def stats(band):
...
+        if values is None:
+            _log.warning(
+                f"save_result: a feature_id_property '{feature_id_property}' was specified,"
+                f" but could not find it in the vector cube: {self._regions}."
+            )
+            return None
+        return dict(enumerate(values))
+
+    def _feature_id_column_name(self) -> Optional[str]:
+        return self.options.get("feature_id_property")  # default would be "id"
```
soxofaan: >  default would be "id"

that's a confusing comment: the default of the implementation is `None`.
What does "would be" mean: a suggestion for future work, or some original behavior that is overruled here?

openeo_driver/save_result.py
```diff
@@ -781,14 +781,38 @@ def stats(band):
...
+
+    def to_csv(self, destination: Union[str, Path, None] = None) -> Union[str, Path, None]:
```
soxofaan, line 807: return None should not happen I think?

openeo_driver/save_result.py
```diff
@@ -781,14 +781,38 @@ def stats(band):
             return self.to_covjson()
         return self.data
 
-    def to_csv(self, destination=None):
+    def _feature_id_mapping(self) -> Optional[dict]:
```
soxofaan, line 784: design-wise it feels there should be just a single helper here instead of two:
`_feature_id_mapping` and `_feature_id_column_name` are highly intertwined I think

soxofaan (commented):
some cosmetic notes,
but looks fine to me

openeo_driver/save_result.py
EmileSonneveld: For me it would be better to throw an exception, but existing code uses `return` without something, so None is possible

tests/test_save_result_csv.py
EmileSonneveld: removed

openeo_driver/save_result.py
EmileSonneveld: clarified with better comment "default would be "id" if the feature needs to be enabled by default"

openeo_driver/save_result.py
EmileSonneveld: They are intertwined, but I don't see a cleaner way. Attempt here https://github.com/Open-EO/openeo-python-driver/commit/d388df37aa0336369516cd38a96ebd8152adea3a

EmileSonneveld:
Example run: https://openeo.dataspace.copernicus.eu/openeo/1.2/jobs/j-26090208511942f7985f36c382c95ca4/results/YTc3ZDBjMjgtNGE3OC00YzBjLThlODUtOTMyNjQzYWQ5ZjU2/4cbce594419c44b85477799c47c201b8?expires=1788947508
```csv
date,feature_index,B04,id
2019-10-15T00:00:00.000Z,1,5947.556842063653,f_middle
2019-10-15T00:00:00.000Z,2,5947.556842063653,f_south
2019-10-15T00:00:00.000Z,0,1345.49498953358,f_north
```

PR #532, opened 2026-09-07
Fix `apply` process not resolving parameters from an enclosing user-d…

EmileSonneveld:
…efined process (e.g. parameterized offset/threshold math) inside its callback. https://github.com/Open-EO/openeo-geopyspark-driver/issues/1739
Written with copilot. Needs review

### DIFF
--- a/openeo_driver/dry_run.py
+++ b/openeo_driver/dry_run.py
@@ -80,38 +80,6 @@ from openeo_driver.utils import EvalEnv, to_hashable
 
 _log = logging.getLogger(__name__)
 
-
-def _resolve_property_filters_with_env(properties: dict, env: EvalEnv) -> dict:
-    """
-    Walk property filter process graphs and replace ``{"from_parameter": X}``
-    references with their literal values from *env*, where X is a UDP-level
-    parameter (not the callback ``"value"`` parameter).
-
-    This prevents "Unknown parameter X" errors later when the property filters
-    are re-evaluated in ``post_dry_run`` without the original UDP parameters.
-    """
-    return properties
-    import copy
-
-    env_params = env.collect_parameters()
-    if not env_params:
-        return properties
-
-    def _substitute(obj):
-        if isinstance(obj, dict):
-            if set(obj.keys()) == {"from_parameter"}:
-                param_name = obj["from_parameter"]
-                # "value" is the callback parameter; leave it as-is
-                if param_name != "value" and param_name in env_params:
-                    return env_params[param_name]
-            return {k: _substitute(v) for k, v in obj.items()}
-        elif isinstance(obj, list):
-            return [_substitute(item) for item in obj]
-        return obj
-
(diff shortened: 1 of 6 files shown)


openeo_driver/processgraph/process_implementations/cubes.py
```diff
@@ -103,6 +104,11 @@ def _contains_only_polygons(geojson: dict) -> bool:
 def apply(args: ProcessArgs, env: EvalEnv) -> DriverDataCube:
     data_cube = args.get_required("data", expected_type=DriverDataCube)
     apply_pg = args.get_deep("process", "process_graph", expected_type=dict)
+    # Substitute parameters already bound in the current environment (e.g. from an enclosing
+    # user-defined process), leaving the callback's own "x"/"context" parameters untouched
+    # (those get resolved later while evaluating the callback itself).
+    # See https://github.com/Open-EO/openeo-geopyspark-driver/issues/1739
+    apply_pg = resolve_child_parameters(apply_pg, env=env, exclude={"x", "context"})
```
EmileSonneveld, line 107: @soxofaan I find this `exclude` suspicious. Does it make sense to you?

soxofaan: that exclude looks fine indeed: the child callback arguments "x" and "context" indeed come directly from it's parent driver "apply", and should not be substituted here

openeo_driver/dry_run.py
```diff
@@ -81,6 +81,37 @@
 _log = logging.getLogger(__name__)
 
 
+def _resolve_property_filters_with_env(properties: dict, env: EvalEnv) -> dict:
```
soxofaan: why is this added? It's not used somehwere or am I missing something?

openeo_driver/processgraph/process_implementations/cubes.py
```diff
@@ -103,6 +104,11 @@ def _contains_only_polygons(geojson: dict) -> bool:
 def apply(args: ProcessArgs, env: EvalEnv) -> DriverDataCube:
```
soxofaan, line 104: apply is not the only process that should be "fixed" here: all processes that take child callbacks should be covered (apply_dimension, reduce_dimension, ...)

soxofaan (commented):
Overall I'm not completely sure this should be "fixed" by just substituting parameters in openeo-python-driver. 

What is also missing is falling back on the default value of a parameter

openeo_driver/dry_run.py
EmileSonneveld: curious, removed.

openeo_driver/processgraph/process_implementations/cubes.py
EmileSonneveld: Added there too

EmileSonneveld:
@soxofaan Default parameters seem to be already filled in when returning `env.collect_parameters()`

PR #538, opened 2026-09-23
feature: EJR: set_status allow specifying an infra_id

pvbouwel:
Currently EJR tracks a backend_id which is a grouping across infrastructure execution environments. This allows to make sure users can see jobs regardless of the backend on which they run but for operational tasks can be improved if the exact infrastructure where a job has run is tracked.


openeo_driver/jobregistry.py
```diff
@@ -109,6 +109,7 @@ def set_status(
         updated: Optional[str] = None,
         started: Optional[str] = None,
         finished: Optional[str] = None,
+        infra_id: Optional[str] = None,
```
soxofaan: isn't "infra_id" an aspect of creation?
I don't think it's intended that the infra_id can be changed with each status update (which this change seems to allow)

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and start a new "In Progress" section above it.
 - Extract openeo-geopyspark-driver oriented batch job result metadata handling from generic view layer and allow openeo-aggregator to inject an alternative implementation in context of "STAC 1.1" batch job result metadata style ([Open-EO/openeo-aggregator/204](https://github.com/Open-EO/openeo-aggregator/issues/204))
 - Fix `save_result` leaking when using UDP sub-evaluation ([#534](https://github.com/Open-EO/openeo-python-driver/issues/534))
 - Add initial support for `reduce_spatial` ([Open-EO/openeo-geopyspark-driver#1715](https://github.com/Open-EO/openeo-geopyspark-driver/issues/1715))
+- Allow specifying `infra_id` when setting a job status
```
soxofaan:
```suggestion
- Add support for `infra_id` in `JobRegistryInterface`
```

